PR #29, opened 2013-09-23
fixes #2961 - Refactoring of output and formatters

mbacovsky:
Foreman side of output refactoring
- shortened field types names
- foreman specific fields and formatters moved


lib/hammer_cli_foreman/output/fields.rb
```diff
@@ -0,0 +1,11 @@
+require 'hammer_cli'
+
+module Fields
```
pitr-ch: Better not to pollute global space, it should be under `HammerCLIForeman`.

PR #153, opened 2014-10-09
Fixes #2928 - support for smart variables and override values

mbacovsky:
- fixed options for smart class parameters
- added commands for manipulating smart-variables
- added override values to smart variables and smart class parameters
- added name options where posiible
- updated test data to Foreman 1.7 develop

Ideas for refactoring are welcome, needs https://github.com/theforeman/foreman/pull/1835 to have the params defined.


mbacovsky:
Rebased...

lib/hammer_cli_foreman/smart_class_parameter.rb
```diff
@@ -97,8 +111,53 @@ class UpdateCommand < HammerCLIForeman::UpdateCommand
...
+    end
+
+    class AddOverrideValueCommand < HammerCLIForeman::CreateCommand
+      resource :override_values
+      command_name 'add-override-value'
+
+      success_message _("Override value created")
+      failure_message _("Could not create the override_value")
+
+      build_options do |o|
```
unorthodoxgeek: nitpick - meaningless variable names are frowned upon in Ruby standards, should be called option instead of o

unorthodoxgeek:
I think this can be merged after the small styling comment. Thanks @mbacovsky :)

mbacovsky:
Thanks for the review @unorthodoxgeek, Styling fixed.

unorthodoxgeek:
@domcleal @ohadlevy - this is good to merge IMHO.

jcmcken:
This possibly just the way I've installed it, but it looks like some options are missing from ``sc-param`` (specifically a way to set the parameter name). Here's some sanitized output from my dev machine:

```console
[root@localhost ~]# hammer --config /etc/hammer/cli.modules.d/login.yml --csv sc-param add-override-value --help
Usage:
    hammer sc-param add-override-value [OPTIONS]

Options:
 --match MATCH                      
 --puppet-class PUPPET_CLASS_NAME  Puppet class name
 --puppet-class-id PUPPET_CLASS_ID ID of puppetclass
 --value VALUE                      
 -h, --help                        print help
```

I see ``--smart-class-parameter`` and --smart-class-parameter-id`` in the tests, though, which leads me to think it's my issue.

mbacovsky:
@jcmcken, this is interesting. Could you check you use recent hammer_cli? DId you regenerate the apipie cache on the server (`bundle exec rake apipie:cache`) and client (run with -r )? 
Help on my setup follows: 
```
Usage:
    hammer sc-param add-override-value [OPTIONS]

Options:
 --match MATCH                                        
 --puppet-class PUPPET_CLASS_NAME                    Puppet class name
 --puppet-class-id PUPPET_CLASS_ID                   ID of Puppet class
 --smart-class-parameter SMART_CLASS_PARAMETER_NAME  Smart class parameter name
 --smart-class-parameter-id SMART_CLASS_PARAMETER_ID  
 --use-puppet-default USE_PUPPET_DEFAULT             One of true/false, yes/no, 1/0.
 --value VALUE                                        
 -h, --help                                          print help
```

jcmcken:
@mbacovsky Hmm, no dice. Is there some dependency on Foreman ``>= 1.7``? We're running ``1.6.x`` in our lab at the moment (``1.6.1`` I think).

 I'm running hammer_cli from the nightly RPM repo (``rubygem-hammer_cli-0.1.4-1.201501260908gitf44b8c5.el6.noarch``). I also have a few other deps from that repo (e.g. ``apipie_binding``). For the Foreman hammer, I built the gem from  	
c7ecffe and installed it.

mbacovsky:
@jcmcken, ahh right, it requires https://github.com/theforeman/foreman/pull/1835 on the server (see PR description). The patch should be applicable even on 1.6.x.

jcmcken:
Ah, well that's unfortunate. Was hoping I could push out an updated hammer to prod without updating other deps. The patch from theforeman/foreman#1835 doesn't seem to apply cleanly to ``1.6.1``, probably missing some intermediate patches somewhere along the line. I'll try getting a dev instance from 151daeb up and running to test

tstrachota:
[test]

tstrachota:
@mbacovsky nice work. I did a bit more extensive testing and found some issues though:

1. Apidoc for both smart variable and override values actions deserve param descriptions. Some parameters have wrong types (`avoid_duplicates`, `merge_overrides`, `use_puppet_default` should be boolean instead of string). I'd deal with it in a separate PR into the API but we need to make sure not to forget about it.
2. In my tests `hammer smart-variable list` displayed significantly less records than UI. Not sure why.
3. Issues in `hammer smart-variable info`
  3.1 Type displays blank value if default is used
  3.2 `Merge overrides` and `Avoid duplicates` flags are missing
  3.3 In some cases override order was displayed blank (while the UI shows it fine)
  3.4 Do we need to nest each override value details under label `Value`? It feels a bit redundant to me when there's already `Values` above all.
  3.5 I wonder if `Count` is useful at all.
4. `hammer  puppet-class smart-variables` - does not puppetclass identifier
5. What is `--use-puppet-default` in `hammer smart-variable add-override-value` good for? I don't see it anywhere in the UI.
6. `hammer smart-variable add-override-value --smart-variable checkout_path` missing value causes exception:
```
Could not create the override_value:
  undefined method `match' for nil:NilClass
```

mbacovsky:
The PR was updated. There is one issue (4) left for tomorrow

1. TBD (file a bug, maybe PR as well)
2. smart variable list is IMO broken in the UI and shows smart class parameters instead
3.1 http://projects.theforeman.org/issues/9811, workaround added
3.2 added
3.3 I was not able to reproduce this, could you elaborate?
3.4 one level removed
3.5 removed
4. it is intentional because your usage does not make much sense (this is actually filtering by puppet class). --puppet-class parameter should be required though (have to figure out how - TBD)
5. UI explanation is 'Do not send this parameter via the ENC. Puppet will use the value defined in the Puppet manifest for this parameter.'  For some reasons I do not recall well in the API this attribute is per override value while in the UI it is on the variable level
6. http://projects.theforeman.org/issues/9812 Not sure if it makes sense to fix this temporarily on the CLI side

Thanks for the rich comments, @tstrachota

mbacovsky:
4. fixed

mbacovsky:
Re 1: the parameter types seem all correct (bool). http://projects.theforeman.org/issues/9823 covers the missing descriptions.

tstrachota:
Ad 3.3: I cannot reproduce it now too but I came across different issue:
1. I created a smart variable via cli `hammer smart-variable create --puppet-class xxx --variable yyy`
2. I removed `hostgroup` line from order field in UI puppet classes > xxx > smart variables > yyy
3. Order in hammer info is broken now:
```
Override values:
, osOrder:  fqdn
```

mbacovsky:
Good catch @tstrachota. Fixed.

lib/hammer_cli_foreman/puppet_class.rb
```diff
@@ -47,11 +48,25 @@ class InfoCommand < HammerCLIForeman::InfoCommand
...
+      build_options_for :puppetclasses
+
+      def validate_options
+        super
+        validator.any(:option_puppetclass_name, :option_puppetclass_id).required
+      end
     end
 
 
+
```
tstrachota: Nitpick: blank line

tstrachota:
@mbacovsky looks and works fine now. Remove the blank line and create a Redmine issue for my comment no. 1, please. I think we can merge then.

mbacovsky:
@tstrachota, thanks for the review. The blank line is removed. The issue for comment no. 1 was created (http://projects.theforeman.org/issues/9823), more info 4 posts above.

tstrachota:
@mbacovsky ok :+1: thanks!

PR #165, opened 2015-01-14
fixes #3630 - adds dns name association to domain cli

cfouant:


lib/hammer_cli_foreman/domain.rb
```diff
@@ -1,5 +1,26 @@
 module HammerCLIForeman
 
+  module DomainUpdateCreateCommons
+
+    def self.included(base)
+      base.option "--dns_id", "DNS_ID",  _("ID of DNS proxy to use within this domain")
```
tstrachota: The parameter should be `--dns-id` to follow conventions.

tstrachota:
@cfouant apart from the parameter name it works fine.

PR #167, opened 2015-02-11
Fixes #9318 - does not resolve a nested host group to id when updating a...

logicminds:
Because nested host groups are not searchable by name (presumable because of the slash in them) we must allow host groups to be searchable via the label attribute.  Without this fix hammer cli foreman does work when trying to either:

a. update a nested host group
b. change the host group of a host to a nested hostgroup

I am not entirely sure why foreman cannot lookup nested hostgroups by name, or if that is a bug in itself.  But this at least allows us to work with nested host groups in the meantime.

Additionally, I added a few more things in the .gitignore file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,8 @@
 Gemfile.lock
 Gemfile.local
 pkg
-
+.bundle
+.idea
 # Tests
 test/reports
 coverage
--- a/lib/hammer_cli_foreman/id_resolver.rb
+++ b/lib/hammer_cli_foreman/id_resolver.rb
@@ -40,7 +40,7 @@ module HammerCLIForeman
       :fact_value =>       [],
       :filter =>           [],
       :host =>             [ s_name(_("Host name")) ],
-      :hostgroup =>        [ s_name(_("Hostgroup name")) ],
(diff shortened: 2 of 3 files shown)


logicminds:
@lzap can we get this merged?

lzap:
I have no push permissions, this is on @orrabin but she is on training this week.

orrabin:
@lzap I don't have push permissions either.

tstrachota:
@logicminds sorry for letting you wait. I got to reviewing the PR now.

I think that `title` is now preferred over `label` in Foreman. Can you please update your patch accordingly? The title (or label if you want) isn't updatable directly, so you should disable it in the resolver:
```ruby
s("title", _("Hostgroup title"), :editable => false)
```
It won't create option `--new-title` in `hammer hostgroup update`

Can you please split your `.gitignore` change and the fix into separate commits?

tstrachota:
[test]

logicminds:
updated as requested.

tstrachota:
Merged as 4d9b9e953e6076f29687ed179b001fd4092456cb and 893d3e7a37f9b029fad4100c51e84d9fd0dcf1b2
Thanks @logicminds !

PR #170, opened 2015-02-22
Refs #8227 - allow disablement of record_to_common_format

mbacovsky:
In some records with only one attribute the heuristic fails.
Having the function linked from command makes the notation shorter

### DIFF
--- a/lib/hammer_cli_foreman/commands.rb
+++ b/lib/hammer_cli_foreman/commands.rb
@@ -176,7 +176,7 @@ module HammerCLIForeman
     end
 
     def send_request
-      HammerCLIForeman.record_to_common_format(super)
+      transform_format(super)
     rescue HammerCLIForeman::MissingSeachOptions => e
 
       switches = self.class.find_options(:referenced_resource => e.resource.singular_name).map(&:long_switch)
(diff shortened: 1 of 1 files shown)


tstrachota:
[test]

tstrachota:
Seems to work well. I have one proposal though:
Could you define a more generic method `transform_data` on `HammerCLIForeman::Command` and give it separate implementation for `ListCommand`? I think that subcommands can do various transformations, not only `*_to_common_format`

mbacovsky:
@tstrachota, might be a good idea, do you have any use case for that? I'd like to avoid confusion with `extend_data` as purpose of the `*_to_common_format` method is to unify the data format coming from various foreman versions. Would name `to_common_format` for the more generic method work for you?

mbacovsky:
@tstrachota, I've added more generic method called `transform_format` to suggest manipulation just with the format of the response. Makes sense?

tstrachota:
@mbacovsky it does, thanks. Do we need to have `*_to_common_format` defined in the command class then?

mbacovsky:
@tstrachota, not necessarily, removed.

tstrachota:
:+1: thanks @mbacovsky :moyai:

PR #171, opened 2015-02-23
refs #8980 - update to gettext 3.x

domcleal:
3.1.13 adds an intermediate .edit.po file alongside each .po, which is meant
to be kept outside of SCM and updated by users, whereupon it's merged back into
the .po on the next rake gettext:find execution.

Previously we overwrote all gettext .po file changes with files directly from
Transifex, but now these are downloaded to .edit.po and gettext merges them
back in.

Continuation of https://github.com/theforeman/hammer-cli/pull/160

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,7 @@ coverage
 
 # Editors
 tags
+
+# Locale files
+locale/*/*.edit.po
+locale/*/*.po.time_stamp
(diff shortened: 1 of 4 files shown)


tstrachota:
[test]

tstrachota:
:+1: thanks @domcleal

PR #172, opened 2015-03-12
Fixes #7699 - can't convert nil into Array in compute resouce info

tstrachota:


mbacovsky:
:+1: Thanks @tstrachota

tstrachota:
Please wait with merging this one. I think it deserves more care. Info on RHEV should print the same columns as on oVirt.

tstrachota:
@mbacovsky I looked at the problem again and updated this PR to what I believe is the right solution. There's also server part of the fix https://github.com/theforeman/foreman/pull/2245

mbacovsky:
@tstrachota, I like this solution much more. Does it makes sense to add `specific_fields` also for RHEV?

mbacovsky:
:+1: I didin't noticed that the related Foreman patch will make addition of RHEV unnecessary.

PR #173, opened 2015-03-16
refs #9648 - use correct domain for system locales, only load one domain

domcleal:
This reverts commit 96ee3d8cb0f0da278bac50ceb4cc15e9092e382f.

Depends on https://github.com/theforeman/hammer-cli/pull/166


mmoll:
Tested on Debian and FreeBSD, works for me. :+1:

mbacovsky:
:+1: Tested also on Fedora. Thanks @domcleal

PR #175, opened 2015-03-31
Fixes #9971 - improve handling of id search errors

mbacovsky:
Do not hide search failures when intentionaly looking for non-required resource id

### DIFF
--- a/lib/hammer_cli_foreman/commands.rb
+++ b/lib/hammer_cli_foreman/commands.rb
@@ -148,7 +148,10 @@ module HammerCLIForeman
       begin
         resolver.send("#{resource.singular_name}_id", opts)
       rescue HammerCLIForeman::MissingSeachOptions => e
-        raise e unless (options[:required] == false)
+        if (options[:required] == true || resource_search_requested(resource, opts))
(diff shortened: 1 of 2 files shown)


mbacovsky:
Successful BATs tests run: http://ci.theforeman.org/job/systest_foreman_hammer_nightly/106/

lib/hammer_cli_foreman/commands.rb
```diff
@@ -157,7 +160,10 @@ def get_resource_ids(resource, options={})
       begin
         resolver.send("#{resource.singular_name}_ids", opts)
       rescue HammerCLIForeman::MissingSeachOptions => e
-        raise e unless (options[:required] == false)
+        if (options[:required] == true || resource_search_requested(resource, opts, plural=true))
```
tstrachota: Please replace `plural=true` with 'true' only. It behaves a bit differently from what I think you expect.

tstrachota:
LGTM, one note in the code. Other than that :+1:

mbacovsky:
Concerned code fixed, thanks for the review, @tstrachota

PR #179, opened 2015-05-19
Fixes #10534 - turn off pagination by default

tstrachota:
Turns off pagination for all list commands by default.
It can still be invoked by passing --per-page explicitly.

### DIFF
--- a/lib/hammer_cli_foreman/commands.rb
+++ b/lib/hammer_cli_foreman/commands.rb
@@ -264,6 +264,7 @@ module HammerCLIForeman
 
     action :index
 
+    RETRIEVE_ALL_PER_PAGE = 1000
     DEFAULT_PER_PAGE = 20
 
     def adapter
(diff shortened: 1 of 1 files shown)


mbacovsky:
It looks good. Loading big data in chunks is a good addition and works nice.
I have two usability concerns:
- if I specify `--page 3 --per-page 20` I'm not sure if I want be asked to go to next page
- the `ui > per_page` variable from hammer config is ignored and Id prefer to use it as a default for the per_page param. So that I can just ask for --page if I have my favorite per_page preconfigured.

tstrachota:
The intended behaviour is:
- no option is passed (default) -> load all internally and print everything
- `--per-page` is passed -> show first page and ask if the user wants to display the next page
- `--par-page` and `--page` is passed -> show the required page and  ask if the user wants to display the next page

It probably makes sense to change the last case not to ask for next page. I can update that.

Anyway I'm afraid there's no place for `per_page` setting when we list all by default. It needs to be removed by a separate commit into `hammer_cli` core.

tstrachota:
Code updated. After thinking about the workflow more I decided to remove the interactive mode completely. It doesn't make much sense keeping it there now.

How it behaves now:
- no option is passed (default) -> load all internally and print everything
- only `--per-page` is passed -> show first page
- only `--page` is passed -> show N-th page, taking value for `per_page` from settings or hardcoded default
- `--per-page` and `--page` is passed -> show the required page and amount

mbacovsky:
This makes much better sense. I tried to experiment with the default per_page size (one of the suggestions in the BZ). `DEFAULT_PER_PAGE = `tput lines`.to_i - 5` worked best for me, but I'm not sure if I prefer it over the fixed size.

LGTM now. Thanks @tstrachota 

:+1:  :snowman:

PR #181, opened 2015-05-27
host_create docs usage is out of date with 0.2.0 release

stevenschlansker:
I'm a little confused as to why these options are in the docs but not the actual code.  A lot of them seem very important to set.  Maybe a better fix is to implement the missing options?  But it is very misleading to document something that does not exist...

### DIFF
--- a/doc/host_create.md
+++ b/doc/host_create.md
@@ -15,15 +15,12 @@ Options:
  --ask-root-password ASK_ROOT_PW           One of true/false, yes/no, 1/0.
  --build BUILD                             One of true/false, yes/no, 1/0.
                                            Default: "true"
- --comment COMMENT                         Additional information about this host
  --compute-attributes COMPUTE_ATTRS        Compute resource attributes.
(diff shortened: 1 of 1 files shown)


tstrachota:
@stevenschlansker you may see different options when you run the same hammer instance against foreman servers with different versions. Most of the options is generated live from the api documentation to reflect features that are available.

tstrachota:
@stevenschlansker I checked the help output and it is correct for `hammer-cli-foreman 0.2.0` and `foreman 1.8`. I created a PR with warning about this https://github.com/theforeman/hammer-cli-foreman/pull/185

@mbacovsky what do you think?

mbacovsky:
@tstrachota I agree the version published corresponds to currently most recent stable versions of Foreman and hammer. The dynamic nature of CLI options may be cause of confusion here so merging #185 sounds like a good idea.

mbacovsky:
Closing this PR in favour of #185. Thanks for the effort and pointing this problem out, @stevenschlansker.

PR #182, opened 2015-05-28
Docs - update example for host create

tstrachota:
The example uses libvirt compute resource. I added a note to emphasize
that attributes for other compute resources are different.


mmoll:
:+1:

PR #183, opened 2015-06-09
Fixes #3968 - adding a command for building PXE default

tstrachota:
- adds `template build-pxe-default`
- extends exception handling to accept messages being stored in the first level of the response hash


mbacovsky:
The code looks well. However if the call ends with `422` it returns response with unexpected format and the hammer error handling fail:
```
{
    "message" => "No TFTP proxies defined, can't continue"
}
```
This can be reproduced when there are no TFTP proxies defined as the message suggests.

tstrachota:
@mbacovsky updated to fix the 422 message issue.

mbacovsky:
:+1: Now it working fine. Thanks, @tstrachota !

PR #186, opened 2015-07-01
Fixes #8015 - Ability to use user's default Org/Location in hammer CLI

alongoldboim:
I have added the ability to set default values for location/organization options for the API commands.
The way you set them is by using Foreman.yml.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@ Gemfile.lock
 Gemfile.local
 pkg
 .bundle
-
+.rvmrc
 # RubyMine
 .idea
 
@@ -21,3 +21,5 @@ tags
 # Locale files
 locale/*/*.edit.po
 locale/*/*.po.time_stamp
+config/cli_config.yml
+config/cli.modules.d/
--- a/config/foreman.yml
+++ b/config/foreman.yml
@@ -14,3 +14,9 @@
 
   # API request timeout. Set to -1 for no timeout
   #:request_timeout: 120 #seconds
+
+  # Set default values for those options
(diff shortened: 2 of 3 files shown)


alongoldboim:
do not review yet. changing stuff.

domcleal:
I don't know if it's what you're changing, but FYI, the current implementation misses the point I think.  Users may have a default org/loc defined in their account on the Foreman, if the feature's enabled, which should be possible to use.  As per my comment on the ticket, some sort of API support would be most useful.

alongoldboim:
I know, i closed this because we decided to implement it on the hammer-cli.
I'm giving to option to define a config file with :global and :resource(e.g user,host) scopes for defining defaults that will be entered to the request if possible.
About the foreman default org/loc, I'm gonna add the option of "Generate defaults from plugin",
every plugin would be able to give hammer a hash and hammer will create a defaults section for it, after building that ill open another pr of adding the implementation for it in hammer-cli-foreman.
sounds ok?

domcleal:
I don't really follow, it doesn't sound related to the original bug report.

alongoldboim:
I'm adding a feature that will enable the user to use his defaults.

mbacovsky:
@domcleal, from what I know this should be CLI's version of the org/location switcher in the UI. . Similiraly this feature should be a way to temporarily switch the org being used (and avoid typing it with every command). Using the user's default org defined on the server should be one of the options.

alongoldboim:
We will have 3 ways to define the defaults, generating from the ui, use the cli to define a defualt, write in the config file.

PR #190, opened 2015-08-18
Fixes #11402 - Add defaults support for location/organization

alongoldboim:
This is the first experiment with the defaults feature.
For now only default organization/location is working.

### DIFF
--- a/lib/hammer_cli_foreman/defaults.rb
+++ b/lib/hammer_cli_foreman/defaults.rb
@@ -1,0 +1,29 @@
+require 'hammer_cli'
+module HammerCLIForeman
+  class Defaults < HammerCLI::BaseDefaultsProvider
+    def initialize
+      @provider_name = 'foreman'
+      @supported_defaults = [:organization_id, :location_id]
+      @description = _('Use the default organization and/or location from the server')
+    end
+
+    def get_defaults(param)
+      param = "default_organization" if param == :organization_id
+      param = "default_location" if param == :location_id
+      user = get_user
+      val = nil
+      if user
+        val = user["results"].first[param] if user["results"]
+        val = val["id"] if val.is_a?(Hash) && param.include?("default")
+      end
+      val
+    end
+
+    private
+    def get_user
+      HammerCLIForeman.foreman_resource(:users).action(:index).call(:search =>"login="+HammerCLIForeman.credentials.username)
+    end
+  end
+  HammerCLI.defaults.register_provider(Defaults.new())
+
+end
--- a/test/unit/defaults_test.rb
+++ b/test/unit/defaults_test.rb
@@ -1,0 +1,32 @@
+require File.join(File.dirname(__FILE__), 'test_helper')
+describe HammerCLIForeman::Defaults do
+
+  context "Defaults" do
+
+    defaults_provider = HammerCLIForeman::Defaults.new
+    user = {"results" => ["default_organization" => {"id" => 2}, "default_location" => {"id" => 1}]}
+    empty_user = {"results" => ["default_organization" => nil, "default_location" => nil]}
+
+    it "returns defaults organization when exisits " do
+      defaults_provider.stubs(:get_user).returns user
+      assert_equal 2, defaults_provider.get_defaults(:organization_id)
+    end
+
+    it "returns nil when defaults organization doesn't exisits " do
+      defaults_provider.stubs(:get_user).returns empty_user
+      assert_nil defaults_provider.get_defaults(:organization_id)
+    end
+
+    it "returns defaults location when exisits " do
+      defaults_provider.stubs(:get_user).returns user
+      assert_equal 1, defaults_provider.get_defaults(:location_id)
+    end
+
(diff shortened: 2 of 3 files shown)


alongoldboim:
Fails because of a change that depends on a change in the hammer_cli that hasn't entered yet.

lib/hammer_cli_foreman/defaults.rb
```diff
@@ -0,0 +1,29 @@
...
+
+module HammerCLIForeman
+
+  class Defaults < HammerCLI::Defaults
+    SUPPORTED_DEFAULTS=[:organization_id, :location_id]
+    def self.register_provider
+      super(self)
+    end
+
+    def self.support?(param)
```
tstrachota: :+1: nice pattern!

lib/hammer_cli_foreman/defaults.rb
```diff
@@ -0,0 +1,27 @@
...
+      SUPPORTED_DEFAULTS.any? {|s| s.to_s == param}
+    end
+
+    def self.get_defaults(param)
+      param = "default_organization" if param == :organization_id
+      param = "default_location" if param == :location_id
+      user = HammerCLIForeman.foreman_resource(:users).action(:index).call(:search =>"login="+HammerCLI::Settings.settings[:foreman][:username])
+      user_id = user["results"][0]["id"] if user["results"]
+      values = HammerCLIForeman.foreman_resource(:users).action(:show).call(:id => user_id)
```
tstrachota: Default organization and location are returned already in the index call so there's probably no need to do the second call here.

lib/hammer_cli_foreman/defaults.rb
```diff
@@ -0,0 +1,27 @@
+require 'hammer_cli'
+module HammerCLIForeman
+
+  class Defaults < HammerCLI::Defaults
```
tstrachota: Do we need to inherit from the main defaults manager class? I don't see much connection here.

alongoldboim: I implemented it that way before we changed to register providers, now we don't really need it.
ill refactor it and update you.
Thanks.

alongoldboim:
@mbacovsky @tstrachota we need to think what we want to implement here and what defaults we want to support, this plugin will serve as a base template for other plugins we would like to serve defaults from.
For now its workingl for org/loc.

alongoldboim:
Another issue we have is that the server calls wont be able to be answered unless user password are set on the conf file.

lib/hammer_cli_foreman/defaults.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+
+    def self.supported_defaults
+      SUPPORTED_DEFAULTS
+    end
+
+    def self.get_defaults(param)
+      param = "default_organization" if param == :organization_id
+      param = "default_location" if param == :location_id
+      user = HammerCLIForeman.foreman_resource(:users).action(:index).call(:search =>"login="+HammerCLI::Settings.settings[:foreman][:username]) if HammerCLI::Settings.settings[:foreman][:username]
```
mbacovsky: To address the issue with missing username you can use HammerCLIForeman.credentials that ask interactively if necessary:
```ruby
      user = HammerCLIForeman.foreman_resource(:users).action(:index).call(:search =>"login="+HammerCLIForeman.credentials.username)
```

mbacovsky:
I think for now organization and location are good to start with. These are the only ones requested in the BZ, we can add more later according to demand.

lib/hammer_cli_foreman/defaults.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+
+    private
+    def get_user
+      HammerCLIForeman.foreman_resource(:users).action(:index).call(:search =>"login="+HammerCLIForeman.credentials.username)
+    end
+  end
+  HammerCLI.defaults.register_provider(Defaults.new())
+
+end
```
mbacovsky: no newline at the end of file

test/unit/defaults_test.rb
```diff
@@ -0,0 +1,32 @@
+require File.join(File.dirname(__FILE__), 'test_helper')
+describe HammerCLIForeman::Defaults do
+
+  context "Defaults" do
+
+    defaults_provider = HammerCLIForeman::Defaults.new
+    user = {"results" => ["default_organization" => {"id" => 1}, "default_location" => {"id" => 1}]}
```
mbacovsky: could you use different id values to avoid possibility overseeing of swapping them by mistake?

mbacovsky:
Besides the little things mentioned above it looks good. 
Could you please rebase?

mbacovsky:
:+1: Good job, @alongoldboim! :beers:

tstrachota:
@alongoldboim great job! I appreciate the effort, guys! :hatched_chick:

alongoldboim:
@tstrachota @mbacovsky mazal tov to all of us :) toke some time, but i like the new feature.

PR #199, opened 2015-10-13
Fixes #12112 - Removing credentials in case login was unsuccessful in interactive mode

alongoldboim:
A change to apipie-bindings was also made to support this patch and must be merged before this one works.(https://github.com/Apipie/apipie-bindings/pull/31)

### DIFF
--- a/lib/hammer_cli_foreman/commands.rb
+++ b/lib/hammer_cli_foreman/commands.rb
@@ -202,6 +202,9 @@ module HammerCLIForeman
         },
         e.resource
       )
+    rescue Exception => e
+      HammerCLIForeman.foreman_api_connection.api.clear_credentials if e.to_s.include?"401"
+      raise e
     end
 
     def transform_format(data)


alongoldboim:
This PR (https://github.com/Apipie/apipie-bindings/pull/31) need to get in

alongoldboim:
@mbacovsky ready to be pushed.

lib/hammer_cli_foreman/commands.rb
```diff
@@ -202,6 +202,9 @@ def send_request
         },
         e.resource
       )
+    rescue Exception => e
+      HammerCLIForeman.foreman_api_connection.api.clear_credentials if e.to_s.include?"401"
```
mbacovsky: `.include?('401')`

mbacovsky: While I think it over again it would be better to catch RestClient::Unauthorized and preferably do it in Hammer core (HammerCLI::Apipie::Command). What do you think?

mbacovsky: Or even - would it make sense to catch the error in apipie-bindings while we do not touch any other Hammer objects?

alongoldboim: if we would like to implement this in apipie, where should i catch the error and clear the credentials?

mbacovsky:
I'd suggest https://github.com/Apipie/apipie-bindings/blob/master/lib/apipie_bindings/api.rb#L222. But he question remains - is it a good idea or you've had a reason to do this in hammer-cli-foreman?

alongoldboim:
I think this could be a default behavior anyway, no need to save bad credentials better clearing them away, and we can do it in both places, but apipie makes more sense when i think it over.

PR #200, opened 2015-10-14
Fixes #11074 - Added IPAM desc on subnet info command.

alongoldboim:


tstrachota:
:+1: thanks @alongoldboim

PR #201, opened 2015-10-19
Fixes #12202 - string parameters get double-quoted

tstrachota:
With this patch
```
hammer host update --id 1 --parameters 'password=secret'
```
no longer creates parameter `password` with double quoted value `'"secret"'`. The value is created without the quotes now.


mbacovsky:
:+1: LGTM. Thanks, @tstrachota

PR #205, opened 2015-12-06
Fixes #7412 - Addng info command to role

orrabin:


tstrachota:
Thanks @orrabin. Can you please extend the response data to print the builtin value as human readable? Something like:
```ruby
def extend_data(data)
  data['builtin_text'] = (data['builtin'].to_i > 0) ? _('Yes') : _('No')
  data
end
```

I've also been thinking, regardless what the BZ says, whether it makes sense to create the info command just to provide the builtin flag. Wouldn't it be better to print it in the `role list`? Or do we want the info command just because it's kind of standard command layout in hammer? What do you think @orrabin, @mbacovsky ?

orrabin:
@tstrachota I think it should be there mainly as a standard command. There might be a request to add the builtin flag to the list as well and I'd still keep the info command.
In the api you also get a list of filter ids, I didn't add it since there is already a `filters` command for role but maybe they should be combined.

tstrachota:
@orrabin ok, let's keep the info command in place. I agree with having filter listing in the other command.
Can you please fix the data formatting as proposed in my previous comment https://github.com/theforeman/hammer-cli-foreman/pull/205#issuecomment-163164787?

orrabin:
@tstrachota I fixed your comment and rebased.

tstrachota:
Works fine, thank you @orrabin!

PR #206, opened 2015-12-07
Fixes #12699 - Commands for setting parameters at taxonomies

tstrachota:
- parameter commands for organizations and locations
- common foundation for functional tests
- base parameter commands updated to enable for displaying names and values in messages

### DIFF
--- a/test/functional/api_expectations.rb
+++ b/test/functional/api_expectations.rb
@@ -11,9 +11,9 @@ module APIExpectations
       resource, action, params, headers, options = actual_parameters.shift(5)
 
       result = true
-      result = result && (resource.to_s == @expected_resource.to_s) unless @expected_resource.nil?
-      result = result && (action.to_s == @expected_action.to_s) unless @expected_action.nil?
-      result = result && @block.call(params) if @block
+      result &&= (resource.to_s == @expected_resource.to_s) unless @expected_resource.nil?
+      result &&= (action.to_s == @expected_action.to_s) unless @expected_action.nil?
+      result &&= @block.call(params) if @block
       result
     end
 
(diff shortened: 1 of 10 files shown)


tstrachota:
Please note that testing the patch may be affected with http://projects.theforeman.org/issues/12725. UI works fine.

test/functional/api_expectations.rb
```diff
@@ -0,0 +1,65 @@
...
+      @expected_resource = resource
+      @expected_action = action
+      @block = block if block_given?
+    end
+
+    def matches?(actual_parameters)
+      resource, action, params, headers, options = actual_parameters.shift(5)
+
+      result = true
+      result = result && (resource.to_s == @expected_resource.to_s) unless @expected_resource.nil?
```
mbacovsky: nit: `&&=`

mbacovsky:
The params functionality works well.
The `location info` has some issues with params rendering:
In the data received there was `                "parameters" => {}` and the info rendered showed:
```
Hostgroups:         

Parameters:         
     =>
Created at:         2015/05/18 11:53:09
Updated at:         2015/06/17 13:18:10
```
Notice the extra arrow. Do you see the same beahvior?

I very much like the functional tests you've introduced. It has a lot of potential. The Mocha extension is a great idea ;). The tests are well readable.
The only concern I have is @param instance var that is taken over among the tests. I'd prefer to keep it local to the test. Did you have a reason for doing it this way?

Any way great addition overall. I'll take deeper look at the tests tmrw.
:pig_nose:

tstrachota:
@mbacovsky issues fixed.

Regarding the `location info` it's the bug I mentioned in https://github.com/theforeman/hammer-cli-foreman/pull/206#issuecomment-162482637. The info command correctly displays wrong data returned from the API.

mbacovsky:
@tstrachota thanks, LGTM
:+1:

PR #210, opened 2016-01-10
Fixes #12998 - hammer listing the sc-params shows puppetclass

orrabin:
This will only work after: https://github.com/theforeman/foreman/pull/3045 is merged.


tstrachota:
Works well, thanks @orrabin !

PR #213, opened 2016-02-17
Fixes #13758 - Adding parent to taxonomies info command

orrabin:


lib/hammer_cli_foreman/location.rb
```diff
@@ -24,6 +24,7 @@ class InfoCommand < HammerCLIForeman::InfoCommand
       option '--id', 'ID', _("Location numeric id to search by")
 
       output ListCommand.output_definition do
+        field :parent_name, _("Parent")
```
tstrachota: This could as well be:
```ruby
field nil, _("Parent"), Fields::SingleReference, :key => :parent
```
which would enable parent ids to be shown when `--show-ids` is used:
```
Parent:   Default Location (id: 2)
```

tstrachota:
@orrabin please consider using the reference field type. Otherwise :+1:

orrabin:
@tstrachota thanks, I didn't know there was an option to show ids.

mbacovsky:
:+1: Thanks for switching to references @orrabin. Works nice.

PR #216, opened 2016-02-24
Fixes #13830 - names of sc-params are immutable

tstrachota:


mbacovsky:
:+1: Thanks for the test, @tstrachota

PR #217, opened 2016-02-27
Fixes #13926 - Environment and Puppet proxy is not required

ares:


lib/hammer_cli_foreman/host.rb
```diff
@@ -382,9 +382,6 @@ def validate_options
             validator.all(:option_environment_id, :option_architecture_id, :option_domain_id,
                           :option_puppet_proxy_id, :option_operatingsystem_id,
```
iNecas: What about removing the `puppet_proxy_id` requirement as part of this PR?

iNecas: as well as `option_environment_id`, this would help being able to provision without puppet things (although the workaround is using a hostgroup)

ares:
@iNecas like this?

iNecas:
Yes! ACK

tstrachota:
Thanks @ares

PR #219, opened 2016-02-28
Fixes #11880 - Added special method for dealing with puppetclasses

ShimShtein:
Puppetclasses API call is using a different format for the output - it uses a hash instead of an array of results.
Created a method that would deal with puppetclasses output.


ares:
ACK (also tested and works nice)

mbacovsky:
Thanks, merging...

PR #221, opened 2016-03-02
fixes #13966 - newer version of apipie validates types for arrays

tstrachota:
Therefore we no longer can send network interfaces in hashes even though
the server would cope with it.

The PR also fixes api mocking and expectations that used to skip param validations.


domcleal:
Note the commit message is missing a `#` before the ticket number.

tstrachota:
Oops, good catch @domcleal Fixed

tstrachota:
The related apipie-bindings PR has been merged, waiting for a new release now.

iNecas:
Tested and works well, also no code comments. Clean :+1:. Thanks @tstrachota !

PR #222, opened 2016-03-04
fixes #7468 - show auth source name for all auth sources

stbenjam:


lib/hammer_cli_foreman/user.rb
```diff
@@ -26,7 +26,7 @@ class InfoCommand < HammerCLIForeman::InfoCommand
 
       output ListCommand.output_definition do
         field :admin, _("Admin"), Fields::Boolean
-        field :auth_source_internal, _("Authorized by"), Fields::Reference
+        field :auth_source_name, _("Authorized by")
```
tstrachota: Could you please use `field nil, _("Authorized by"), Fields::SingleReference, :key => :auth_source` ?
It will enable for showing the auth source id when `--show-ids` is set.

stbenjam: Oh thanks, I ddin't realize that was a thing.

tstrachota:
:+1: :cake: Thanks @stbenjam

PR #223, opened 2016-03-10
Fixes #11147 - Handle API request redirects with useful message

mbacovsky:
- added dependency on apipie-bindings as we require more strict version than core
- set to not follow redirect by default
- detect redirect from HTTP to HTTPS
- add some hints on how to update the config (does it make sense?)
- smaple config updated

How to test: 
run some queries to sat6 or foreman with enabled redirects with protocol set to http.


tstrachota:
Works nicely, thanks @mbacovsky :golf:  :goat:

PR #224, opened 2016-03-18
Fixes #13927 - make primary and provision flag optional

domcleal:
(cherry picked from commit e8a7110e0699cb54875df67b7d454a687acdcc9d)

---

Proposed cherry pick of #13927 to fix ticket http://projects.theforeman.org/issues/14267, seen in 0.6.1 on Foreman 1.11.

Jenkins isn't configured to run tests on 0.6-stable, sorry about that.


xcompass:
Got the same issue here on foreman 1.10. Tried to debug before I found this. It seem the source of the issue is `nics` passed into `check_mandatory_interfaces` is a simple array. So removing key on line 153 and 156 fixes the problem too.

mbacovsky:
:+1: Thanks @domcleal, tests ran localy are green.

PR #226, opened 2016-03-31
Fixing path in docs for generated test json

jlsherrill:


tstrachota:
:+1: thanks @jlsherrill

PR #227, opened 2016-04-07
Fixes #7492 - Provide success/failure message for associating commands

adamruzicka:


lib/hammer_cli_foreman/commands.rb
```diff
@@ -522,6 +527,22 @@ def self.desc(desc=nil)
       description.strip.empty? ? _("Associate a resource") : description
     end
 
+    def self.failure_message(msg = nil)
+      if msg.nil?
+        super(default_message "Could not associate the %{resource_name}")
```
tstrachota: Translations, please.

lib/hammer_cli_foreman/commands.rb
```diff
@@ -522,6 +527,22 @@ def self.desc(desc=nil)
...
+      else
+        super msg
+      end
+    end
```
tstrachota: We use the massage methods in DSL manner. When no argument is passed we suppose it's called as a getter. Please see implementation of `super` or `self.desc` above. Therefore the overriding method should have IMHO been:
```ruby
def self.failure_message(msg = nil)
  super || default_message(_("Could not associate the %{resource_name}"))
end
```
The current implementation always hides the original messages that were set in the command.

tstrachota:
Thanks, @adamruzicka. I added some comments inline.

Providing some default messages is a good pattern. When it comes to localization, we translate resource names too. Would you mind adding specific messages to into `lib/hammer_cli_foreman/associating_commands.rb` where they are missing?

adamruzicka:
@tstrachota Added the specific messages and fixed usage of the methods as getters

tstrachota:
:+1: thanks @adamruzicka merging!

PR #232, opened 2016-04-29
Document vmware `start` parameter

positron:
In our testing "true" didn't work like the foreman docs show for other REST parameters. `start` must be `"1"` for the VM to start so make these docs extra clear since they are the only docs on the entire internet we could find for this parameter..


tstrachota:
Merged as 1391a555e311f27c66b06356af30716e6a2de443
Thanks @positron :+1:

PR #233, opened 2016-05-04
Fixes #13832 - Forbid setting smart param override value and puppet d…

adamruzicka:
…efault at the same time

This however revealed this http://projects.theforeman.org/issues/14927

Requires:
- [x] https://github.com/theforeman/hammer-cli/pull/207

### DIFF
--- a/lib/hammer_cli_foreman/smart_class_parameter.rb
+++ b/lib/hammer_cli_foreman/smart_class_parameter.rb
@@ -138,6 +138,8 @@ module HammerCLIForeman
       end
 
       validate_options do
+        any(:option_value, :option_use_puppet_default).required
+        all(:option_value, :option_use_puppet_default).rejected
         if option(:option_smart_class_parameter_name).exist?
           any(:option_puppetclass_name, :option_puppetclass_id).required
         end
--- a/lib/hammer_cli_foreman/smart_class_parameter.rb
+++ b/lib/hammer_cli_foreman/smart_class_parameter.rb
@@ -138,8 +138,10 @@ module HammerCLIForeman
       end
 
       validate_options do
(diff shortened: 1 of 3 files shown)


tstrachota:
@adamruzicka beware, the validator checks only for presence of the option, regardless of its value. Therefore `hammer sc-param add-override-value --use-puppet-default=0 --value=VAL` will be accepted as a valid combination.

We could either change the option to flag
```ruby
option '--use-puppet-default', :flag, _('...')
build_options do |options|
  options.without(:smart_variable_id, :use_puppet_default)
  #...
```
or check it using `option_use_puppet_default` in the command's `execute` method. Or add simple support for testing option values in the validator too.

Btw I've had this code for validating `one_of` options in a branch for some time https://github.com/tstrachota/hammer-cli/commit/b649085595dcd06e90cd0ca7fbf7e841072e47b8.
```ruby
one_of(:option_value, :option_use_puppet_default).required
```
I can turn it into a PR if you want (I've changed the commit message to be prepared :smile:).

tstrachota:
@adamruzicka per the discussion we had I opened:
https://github.com/theforeman/hammer-cli/pull/207

You can now do:
```ruby
if option(:option_use_puppet_default).value
  option(:option_value).rejected(:msg => _("You can't set --value when --use-puppet-default is true"))
end
```

tstrachota:
@adamruzicka looks good. Would you mind to add one more test for the case I discovered?
`params = ['--value', 'something', '--use-puppet-default', false]`

tstrachota:
[test]

adamruzicka:
@tstrachota the tests are green now

tstrachota:
Merged, thanks @adamruzicka

PR #235, opened 2016-05-05
VMWare >> VMware (2X)

pondrejk:


tstrachota:
:+1: thanks @pondrejk

PR #237, opened 2016-05-07
Fixes #14510 - The resources /reports were moved to /config_reports

ranjan:


ares:
I think it has to changed in `./lib/hammer_cli_foreman/report.rb` as well. Ideally also command name and documentation should be changed as there can be more report types now.

tstrachota:
Correct, @ares. Another option could be to keep it as `hammer host reports` / `hammer report` and add flags like `--config` or `--type=config` for filtering by type.

PR #239, opened 2016-05-09
i18n - remove underscore from override value error message

domcleal:
Reported at https://www.transifex.com/foreman/foreman/translate/#de/hammer-cli-foreman/63251212


tstrachota:
Thanks @domcleal

PR #240, opened 2016-05-12
fixes #4918 - add realm commands to hammer

stbenjam:


stbenjam:
[test]

test/unit/realm_test.rb
```diff
@@ -0,0 +1,87 @@
...
+    end
+
+  end
+
+  context "CreateCommand" do
+
+    let(:cmd) { HammerCLIForeman::Realm::CreateCommand.new("", ctx) }
+
+    context "parameters" do
+      it_should_accept "name, type, proxy", ["--name=realm", "--realm-type=FreeIPA", "--realm-proxy-id=1"]
```
akofink: This should not only accept these parameters but also require them, right? Perhaps writing a test which asserts failure when these options aren't specified would exercise this behavior.

akofink:
The functionality implemented works for me. The tests make sense and pass locally.

Do you think it would be wise to test that the proper API endpoints are being called?

stbenjam:
> Do you think it would be wise to test that the proper API endpoints are being called? 

Is there an example in here?  This PR is literally a cut and paste job from the `domain` commands

akofink:
Yes. They're usually in the functional tests, such as [hosts](https://github.com/theforeman/hammer-cli-foreman/blob/master/test/functional/host_test.rb).

stbenjam:
That seems rather overkill, it's not done for domains, and realms is basically the same.  I would prefer this get into hammer as it's been lacking for some time.  The functionality works and it's tested like everything else is.

lib/hammer_cli_foreman/realm.rb
```diff
@@ -0,0 +1,53 @@
...
+        HammerCLIForeman::References.taxonomies(self)
+        HammerCLIForeman::References.timestamps(self)
+      end
+
+      build_options
+    end
+
+
+    class CreateCommand < HammerCLIForeman::CreateCommand
+      success_message _("Realm [%{name}] created")
```
stbenjam: It is in brackets for domains

test/unit/realm_test.rb
stbenjam: These kinsd of tests are commented out in the others because they don't work, for example, see the user_test: https://github.com/theforeman/hammer-cli-foreman/blob/master/test/unit/user_test.rb#L58-L62

lib/hammer_cli_foreman/realm.rb
```diff
@@ -0,0 +1,53 @@
...
+        field :name, _("Name")
+      end
+
+      build_options
+    end
+
+
+    class InfoCommand < HammerCLIForeman::InfoCommand
+      output ListCommand.output_definition do
+        field :realm_proxy_id, _("Realm proxy id")
```
tstrachota: Just one nitpick: I haven't tested the API, but if it returns `realm_proxy_name` too, it would be good to use
```ruby
field nil, _("Realm Proxy"), Fields::SingleReference, :key => :realm_proxy
```
instead, to show name by default and id when user asks for it (`--show-ids`).

tstrachota: I checked the api and proxy name isn't contained in the show reply at the moment. Disregard my previous post.

mbacovsky:
:+1: Thanks to all involved!

PR #241, opened 2016-05-19
Removing wrong colons from host create docs

tstrachota:


mbacovsky:
:+1:  LGTM, thanks @tstrachota  :scroll:

PR #243, opened 2016-06-06
Fixes #15257 - let print adapters decide whether to paginate

tstrachota:
Requires https://github.com/theforeman/hammer-cli/pull/208


lib/hammer_cli_foreman/commands.rb
```diff
@@ -328,7 +328,9 @@ def pagination_supported?
     end
 
     def should_retrieve_all?
-      pagination_supported? && option_per_page.nil? && option_page.nil? && HammerCLI::Settings.get(:ui, :per_page).nil?
+      retrieve_all = (pagination_supported? && option_per_page.nil? && option_page.nil?)
+      retrieve_all = (retrieve_all && HammerCLI::Settings.get(:ui, :per_page).nil?) if output.adapter.paginate_by_default?
```
akofink: Just a style thing - it could be shortened to:

```ruby
retrieve_all &&= HammerCLI::Settings.get(:ui, :per_page).nil? if output.adapter.paginate_by_default?
```

akofink:
While we're here, do you think 1000 is a good buffer size for any table? https://github.com/theforeman/hammer-cli-foreman/blob/master/lib/hammer_cli_foreman/commands.rb#L272

test/unit/commands_test.rb
```diff
@@ -194,3 +194,87 @@ class TestList < HammerCLIForeman::ListCommand
...
+  end
+
+  def build_items(cnt)
+    (1..cnt).map do |i|
+      {:id => i, :name => "Item #{i}"}
+    end
+  end
+
+  def expect_paged_call(page, per_page, item_cnt)
+    api_expects(:domains, :index, "List records page #{page}") do |par|
```
akofink: All of our other tests which hit the stubbed API are considered functional tests because they test behavior with respect to the API. Usually unit tests test a single method without exercising any of its interfaces. Could you please justify why you've placed this in the unit tests directory?

mbacovsky: I don't have strong preference here but I won't mind leaving it in unit tests as it is tested on abstract command that is not exposed to the user.

mbacovsky:
I liked it @tstrachota, thanks for the tests.
:+1: from me when @akofink is happy with it.

tstrachota:
@akofink regarding the buffer size, you're right that one number probably won't fit all tables, but I'd solve it as a separate issue based on statistics from real world usage rather than guessing the numbers now.

test/unit/commands_test.rb
tstrachota: I was actually hesitating where to put it, but chose unit tests in the end for the reason @mbacovsky mentioned - it's testing behaviour aspect of an abstract command the users don't know about (also without checking what's being printed to the output).
I'm ok to move it if you think it's wrong. In fact, I don't care too much about where the tests are placed as long as they exist ;-)

lib/hammer_cli_foreman/commands.rb
tstrachota: Nice, updated

tstrachota:
@akofink @mbacovsky thanks for the comments! Updated, some replies inline.

tstrachota:
[test]

mbacovsky:
[test]

test/unit/commands_test.rb
akofink: @tstrachota I'd still call them functional because HammerCLIKatello and other libraries use this abstract class. I mostly noticed the test location because you had to move the API expectation helpers out of the `test/functional/test_helper.rb`. Also, what about the [ListCommand tests that already exist](https://github.com/theforeman/hammer-cli-foreman/blob/master/test/functional/commands/list_test.rb)? It's no show stopper for me, though, if you all agree that they're better classified as unit tests.

mbacovsky:
@tstrachota it seems that recent changes in validators https://github.com/theforeman/hammer-cli/pull/207 changed the output message used and it seems it is the reason the tests fail here.

tstrachota:
[test]

tstrachota:
@mbacovsky a fix has been merged so the tests are green now. I'm going to move the tests and then we can hopefully merge and release 0.7

tstrachota:
@mbacovsky @akofink I moved the tests into functional and merged them with the existing ones. Unfortunately it was extremely slow to check output of the command that printed thousands of records, so I decided to keep those two things separate as "api interaction" and "pagination output".

akofink:
LGTM! ACK

tstrachota:
Thank you, merged

PR #245, opened 2016-06-14
Fixes #15415 - Use master branch for hammer_cli...

akofink:
Use the master branch for hammer_cli in development and test

This would facilitate testing when dependent changes to hammer_cli are made.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,6 +14,10 @@ group :test do
   gem 'ci_reporter', '>= 1.6.3', "< 2.0.0", :require => false
 end
 
+group :development, :test do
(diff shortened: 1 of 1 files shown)


akofink:
Jenkins config needs to be altered to not add hammer_cli in the Gemfile.local

tstrachota:
@akofink can't we just use `Gemfile.local.rb`? See https://github.com/theforeman/hammer-cli-foreman/blob/master/Gemfile#L18-L21 for details.

I don't like the idea of hardcoding the source (even for dev and test environment).

PR #246, opened 2016-06-16
Fixes #15433 - Update documented test data path

akofink:


mbacovsky:
@akofink, thanks for keeping our docs up to date! :beer: 
:+1:

PR #251, opened 2016-07-26
Fixes #15833 - add filtering by env to smart class parameter id resolver

mbacovsky:
When there is the same class imported from multiple environments
its parameters are listed for each of the respective environments.
This patch is adding filtering by --environemnt-id (--environment)
to sc-param update and info commands.


lfisher47:
Adding environment filter solves the problem, but does it really make sense when the smart class lists are not distinct per environment?  Can I really set a default for one environment and not set it for another?

mbacovsky:
@lfisher47, you are right. After looking more deeper into how params work I'm going to address this on the API level.

PR #253, opened 2016-08-05
Fixes #15990 - Hostgroup info should show cv and lifecycle-env

ranjan:


ranjan:
[test] failure looks unrelated to the change

tstrachota:
@ranjan I'm afraid this code belongs rather to hammer_cli_katello, because it's Katello specific. Can you please move your patch to:
https://github.com/Katello/hammer-cli-katello/blob/master/lib/hammer_cli_katello/hostgroup_extensions.rb

Thanks!

ranjan:
thanks @tstrachota, will move this changes to hammer_cli_katello.

PR #254, opened 2016-08-10
Fixes #16016 - Usergroup Inherited Roles showing in hammer

dhlavac:
Added showing inherited usergroups and inherited roles in hammer

### DIFF
--- a/lib/hammer_cli_foreman/references.rb
+++ b/lib/hammer_cli_foreman/references.rb
@@ -31,8 +31,18 @@ module HammerCLIForeman
 
     def self.usergroups(dsl)
       dsl.build do
-        collection :usergroups, _("User groups"), :numbered => false do
+        collection :usergroups, _("User groups"), :numbered => true do
           custom_field Fields::Reference
+          collection :roles, _("Roles"), :numbered => false do
+            custom_field Fields::Reference
+          end
+        end
+        collection :cached_usergroups, _("Inherited User groups"), :numbered => true do
+          field :id, "Id", Fields::Id
+          field :name, "Usergroup"
+          collection :cached_roles, _("Roles"), :numbered => false do
+            custom_field Fields::Reference
+          end
         end
       end
     end
--- a/lib/hammer_cli_foreman/references.rb
+++ b/lib/hammer_cli_foreman/references.rb
@@ -32,15 +32,16 @@ module HammerCLIForeman
     def self.usergroups(dsl)
       dsl.build do
         collection :usergroups, _("User groups"), :numbered => true do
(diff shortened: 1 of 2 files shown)


lib/hammer_cli_foreman/references.rb
```diff
@@ -31,8 +31,19 @@ def self.users(dsl)
 
     def self.usergroups(dsl)
       dsl.build do
-        collection :usergroups, _("User groups"), :numbered => false do
-          custom_field Fields::Reference
+        collection :usergroups, _("User groups"), :numbered => true do
+          field :name, "Usergroup"
+          field :id, "Id", Fields::Id
```
tstrachota: Translations, please.

lib/hammer_cli_foreman/references.rb
```diff
@@ -31,8 +31,19 @@ def self.users(dsl)
...
+          collection :roles, _("Roles"), :numbered => false do
+            custom_field Fields::Reference
+          end
+        end
+        collection :cached_usergroups, _("Inherited User groups"), :numbered => true do
+          field :name, "Usergroup"
+          field :id, "Id", Fields::Id
```
tstrachota: Same here, translation is missing.

lib/hammer_cli_foreman/user.rb
```diff
@@ -43,6 +43,7 @@ def extend_data(user)
         user["locale"] ||= _('default') if user.has_key?("locale")
         user["timezone"] ||= _('default') if user.has_key?("timezone")
         user["full_name"] = [user["firstname"], user["lastname"]].join(' ')
+        user["cached_usergroups"] = user["cached_usergroups"] - user["usergroups"]
```
tstrachota: This gave me `Error: undefined method `-' for nil:NilClass` when I tried to run the command without the server counterpart. Could you please use `user.fetch('cached_usergroups', [])` to keep compatibility with older versions of Foreman?

tstrachota:
Some notes inline, other than that it works fine.

lib/hammer_cli_foreman/user.rb
dhlavac: It doesn't work with `user.fetch('cached_usergroups',` [])` because I need get difference between all user groups and inherited user groups because of duplicity.

tstrachota:
Works fine now, thanks. I'll re-test and merge when the server counterpart is merged https://github.com/theforeman/foreman/pull/3709

dhlavac:
@tstrachota Thanks

tstrachota:
@dhlavac merged, thanks!

PR #255, opened 2016-08-11
Fixes #16059 - displaying use_puppet_default in sc-param info

orrabin:


lib/hammer_cli_foreman/smart_class_parameter.rb
```diff
@@ -74,6 +75,7 @@ class InfoCommand < HammerCLIForeman::InfoCommand
               field :id, _('Id')
               field :match, _('Match')
               field :value, _('Value')
+              field :use_puppet_default, _('Use puppet default')
```
tstrachota: One thing to consider would be using `Fields::Boolean` type to make it print "yes/no" instead of "true/false"

tstrachota:
[test]

orrabin:
@tstrachota thanks, I fixed your comments

tstrachota:
:+1: Thanks @orrabin

PR #256, opened 2016-08-17
Fixes #16119 - Renamed name to variable for smart_variables

ShimShtein:


iNecas:
I've tested it and I can confirm it fixes the issue. However, there is inconsistency now in how we use `name` vs `variable`: for example, calling `hammer smart-variable list` uses `name` in the table header. Also, this changes the way the `smart-variables` command behaves: it used to take `name` attribute before, therefore we should keep the `--name` option there for some time, making it as deprecated for some time: hammer core has a way to deprecated options https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/options/option_definition.rb#L28

tstrachota:
[test]

ShimShtein:
Added back the `--name` option with deprecation warning. Also changed the field to show value instead of name.

ShimShtein:
Forgot one test.

tstrachota:
Looks good @ShimShtein.

One more detail: I think we need to add legacy `--name`with deprecation warning to `smart-variable delete` too.

iNecas:
Tested and works well. I've noticed one behaviour worth addressing, but outside of this PR http://projects.theforeman.org/issues/16161

tstrachota:
@iNecas @ShimShtein FYI it was quite easy addition so I created a PR right away:
https://github.com/theforeman/hammer-cli/pull/212

ShimShtein:
Added delete scenario.

tstrachota:
:+1: thanks @ShimShtein

PR #259, opened 2016-09-26
Fixes #2921 - Add commands for Audits

gnurag:
- Added audit command and tests.

### DIFF
--- a/lib/hammer_cli_foreman/audit.rb
+++ b/lib/hammer_cli_foreman/audit.rb
@@ -1,0 +1,72 @@
+module HammerCLIForeman
+
+  class Audit < HammerCLIForeman::Command
+
+    resource :audits
+
+    class ListCommand < HammerCLIForeman::ListCommand
+
+      def request_params
+        params = super
+        params['search'] ||= 'time >= "1 month ago"'
+        params
+      end
+
+      def self.included(base)
+        base.extend_help do |h|
+          h.section(_('Override audit record pagination')) do
+            h.text(_('Audit records for past one month are returned by default. This behavior can be changed by passing --search option.'))
+          end
+        end
+      end
+
+      output do
+        field :id, _("Id")
+        field :created_at, _("At"), Fields::Date
+        field :remote_address, _("IP")
+        field nil, _("User"), Fields::SingleReference, :key => :user
+        field :action, _("Action")
+        field :auditable_type, _("Audit type")
+        field nil, _("Audit record"), Fields::SingleReference, :key => :auditable
+      end
+
+      build_options
+    end
+
+
+    class InfoCommand < HammerCLIForeman::InfoCommand
+
+      def extend_data(audit)
+        audit["changes"] = audit["audited_changes"].map do |attribute, change|
+          if change.is_a?(Array)
+            {
+              'attribute' => attribute,
+              'old' => (change[0] ? change[0] : change[0].to_s),
+              'new' => change[1]
+            }
+          elsif !change.nil?
+            {
+              'attribute' => attribute,
+              'value' => change
+            }
+          end
+        end.compact
+        audit
+      end
+
+      output ListCommand.output_definition do
+        collection :changes, _("Audited changes") do
+          field :attribute, _("Attribute")
+          field :value, _("Value"), Fields::Field, :hide_blank => true
+          field :old, _("Old"), Fields::LongText, :hide_blank => true
+          field :new, _("New"), Fields::LongText, :hide_blank => true
+        end
+      end
+
+      build_options :without => [:name]
+    end
+
+    autoload_subcommands
+  end
+
+end
--- a/lib/hammer_cli_foreman.rb
+++ b/lib/hammer_cli_foreman.rb
@@ -41,6 +41,10 @@ module HammerCLIForeman
       'HammerCLIForeman::AuthSource', 'hammer_cli_foreman/auth_source'
     )
 
+    HammerCLI::MainCommand.lazy_subcommand('audit', _("Search audit trails."),
+      'HammerCLIForeman::Audit', 'hammer_cli_foreman/audit'
+    )
+
     HammerCLI::MainCommand.lazy_subcommand('compute-resource', _("Manipulate compute resources"),
       'HammerCLIForeman::ComputeResource', 'hammer_cli_foreman/compute_resource'
     )
--- a/lib/hammer_cli_foreman/id_resolver.rb
+++ b/lib/hammer_cli_foreman/id_resolver.rb
@@ -34,6 +34,7 @@ module HammerCLIForeman
 
     SEARCHABLES = {
       :architecture =>     [ s_name(_("Architecture name")) ],
+      :audit =>            [],
       :compute_resource => [ s_name(_("Compute resource name")) ],
       :domain =>           [ s_name(_("Domain name")) ],
       :environment =>      [ s_name(_("Environment name")) ],
--- a/test/functional/audit_test.rb
+++ b/test/functional/audit_test.rb
@@ -1,0 +1,126 @@
+require File.join(File.dirname(__FILE__), 'test_helper')
+
+describe 'audit' do
+  describe 'audit-list' do
+    let (:cmd) { ['audit', 'info'] }
+    let(:params) { ['--id=83'] }
+    let(:audit_changed) do
+      {
+        'id' => 83,
+        'created_at' => "2017-10-31 11:25:44 UTC",
+        'remote_address' => '::1',
+        'user_name' => "foreman_api_admin",
+        'user_id' => 11,
+        'action' => "update",
+        'auditable_type' => "ProvisioningTemplate",
+        'auditable_name' => "default_location_subscribed_hosts",
+        'auditable_id' => 32,
+        'audited_changes' => {
+          'value' => [
+            nil,
+            "--- false\n..."
+          ]
+        }
+      }
+    end
+    let(:audit_new) do
+      {
+        "user_id" => 1,
+        "user_type" => nil,
+        "user_name" => "foreman_admin",
+        "version" => 1,
+        "comment" => nil,
+        "associated_id" => nil,
+        "associated_type" => nil,
+        "remote_address" => "::1",
+        "associated_name" => nil,
+        "created_at" => "2017-10-09 23:42:44 -1100",
+        "id" => 63,
+        "auditable_id" => 5,
+        "auditable_name" => "John Doe",
+        "auditable_type" => "User",
+        "action" => "create",
+        "audited_changes" => {
+          "login" => "john",
+          "firstname" => "John",
+          "lastname" => "Doe",
+          "mail" => "john@ipa.test",
+          "admin" => false,
+          "auth_source_id" => 3,
+          "locale" => nil,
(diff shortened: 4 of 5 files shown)


domcleal:
[test]

lib/hammer_cli_foreman/audit.rb
```diff
@@ -0,0 +1,36 @@
...
+      end
+
+      build_options
+    end
+
+
+    class InfoCommand < HammerCLIForeman::InfoCommand
+
+      output ListCommand.output_definition do
+        HammerCLIForeman::References.parameters(self)
```
tstrachota: There are no parameters for audits. The fields will be always empty.

lib/hammer_cli_foreman/audit.rb
```diff
@@ -0,0 +1,36 @@
...
+        HammerCLIForeman::References.timestamps(self)
```
tstrachota: I think that `created_at` is the only timestamp for audits and we already print it in the `At` field so I'd not include timestamps here.

lib/hammer_cli_foreman/audit.rb
```diff
@@ -0,0 +1,36 @@
...
+    class ListCommand < HammerCLIForeman::ListCommand
+
+      output do
+        field :id, _("Id")
+        field :created_at, _("At")
+        field :remote_address, _("IP")
+        field :user_name, _("User name")
+        field :action, _("Action")
+        field :auditable_type, _("Audit type")
+        field :auditable_name, _("Audit name")
```
tstrachota: Similar here:
```ruby
field nil, _("Audited record"), Fields::SingleReference, :key => :auditable
```
I'm not sure about what's the best label for this field though.

lib/hammer_cli_foreman/audit.rb
```diff
@@ -0,0 +1,36 @@
...
+      end
+
+      build_options
+    end
+
+
+    class InfoCommand < HammerCLIForeman::InfoCommand
```
tstrachota: Currently there's not much added value in the info command as it prints the same as list. I was thinking about providing some more details about the changed fields and values there. Similar to what UI does:
```
Default changed from --- true ... to --- false ...
Value changed from N/A to --- true ...
```
What do you think?

It will probably require some extending/mapping of the data that come form api. Hammer commands have method `extend_data(record)` for that. Let me know in case you need some help with that.

tstrachota:
Nice addition @gnurag, thanks! Output fields need some changes and I proposed one improvement for the info command.

tstrachota:
Hi @gnurag did you have a chance to look at my comments? Do you think you could re-visit the PR and address them?

gnurag:
Hello @tstrachota! Yes, I did read your comments. I'll make the changes and re-submit them. Thank you!

tstrachota:
Hi @gnurag, any progress on the changes?

lib/hammer_cli_foreman/audit.rb
gnurag: I renamed it to "Audit record". Along with "Audit type", I guess it would make sense.

lib/hammer_cli_foreman/audit.rb
gnurag: My initial idea behind `audit info` was to help with audit record entries that may be very large and may not fit properly in the `audit list` output.

But your suggestion makes sense. It will be useful to indicate what changed in this audit record. Can you point me to a snippet I can use as reference?

lib/hammer_cli_foreman/audit.rb
gnurag: Thanks, removed the parameters and timestamps refs. Didn't realize both timestamps were same :)

lib/hammer_cli_foreman/audit.rb
tstrachota: So api returns `audited_changes` for each audit. It's a hash where keys are the changed attributes and values are either a new value or array of [old, new] values, depending on whether it was an update or creation.

For example:
```ruby
"audited_changes" => {
  "name" => "subtest",
  "ignore_types" => [
    "ProvisioningTemplate",
    "Hostgroup"
  ],
  "description" => "",
  "label" => nil,
  "ancestry" => "2"
}
```

Therefore I would say we need to pre-process the api output first, to move values into some more predictable hashes. In hammer you can use method `extend_data` that gets the api data as a parameter. Something like:
```ruby
def extend_data(audit)
  audit["changes"] = audit["audited_changes"].map do |attribute, change|
    if change.is_a?(Array)
      {
        'attribute' => attribute,
        'old' => change[0],
        'new' => change[1]
      }
    elsif !change.nil?
      {
        'attribute' => attribute,
        'value' => change
      }
    end
  end.compact
  audit
end
```
...or some other modification based on how we want the output to look like.

I did some experiments and following output is the best I've been able to achieve so far:
example 1:
```
Audited changes: 
 1) Attribute: name
    Value:     tst
 2) Attribute: ancestry
    Value:     3
 3) Attribute: grub_pass
    Value:
 4) Attribute: lookup_value_matcher
    Value:     hostgroup=tst2/tst
```
example 2:
```
Audited changes: 
 1) Attribute: default
    Old:       
      --- on_demand
      ...
    New:       
      --- immediate
      ...
```

The field definition I used for that is:
```ruby
  collection :changes, _("Audited changes") do
    field :attribute, _("Attribute")
    field :value, _("Value"), Fields::Field, :hide_blank => true
    field :old, _("Old"), Fields::LongText, :hide_blank => true
    field :new, _("New"), Fields::LongText, :hide_blank => true
  end
end
```
I didn't invest too much time into that so there are probably more readable ways of formatting the changes.

mbacovsky:
[test]

ohadlevy:
whats the status on this one? thanks :)

gnurag:
looking into completing this pr.

ares:
@gnurag any progress? I'd love to see this feature in

ares:
Notes from testing

1) it might be related by API but `hammer audit list` ignores pagination and displays all my 6000 audits if I don't specify `--per-page 10` or `--page 1`
2) I see `--organization-id` and related params for both list and info but AFAIK audits are not (yet) linked to any organizations
3) info command has `--name` attribute but using it ends up with 400 `Field 'name' not recognized for searching!`, audits don't have name
4) displaying info a an audit does not seem to display old value, always only the new value (even for update audits)

other than that, it's a great addition

ares:
@gnurag any chance to finish this before 1.18 freeze? Feel free to ignore point 2, audits will receive link to taxonomies soon.

rahulbajaj0509:
I am working on it, will have the pr ready by next week :)

ares:
thanks @rahulbajaj0509, let me know if you need any help

rahulbajaj0509:
About, pagination, should we set some default like 100 records? I am not sure about the pagination as it will break the consistency with other commands. Other option that comes to my mind is that we can add the `--per-page` attribute as a requried parameter. 

What are your thoughts on this? @ares  @tstrachota

ares:
I didn't know back then that we are setting per page to 9999 by default, if this is how it behaves everywhere else, I'd keep it consistent. And hope that we'll set reasonable per-page value globally in short future.

tstrachota:
At the moment hammer lists all the records...

gnurag:
Audit records returned by `hammer audit list` are now limited to last 1 month. Also added a note in `hammer audit list --help` explaining this.

lib/hammer_cli_foreman/audit.rb
```diff
@@ -0,0 +1,76 @@
...
+          h.section(_('Override audit record pagination')) do
+            h.text(_('Audit records for past one month are returned by default. This can be paginated further by using --per-page and --page options.'))
+          end
+        end
+      end
+
+    end
+
+    class ListCommand < HammerCLIForeman::ListCommand
+      include PerPageAuditLimit
```
tstrachota: If this is the only place where we're going to use the module, I'd vote for deleting the module and just moving the logic to this class.

lib/hammer_cli_foreman/audit.rb
```diff
@@ -0,0 +1,76 @@
...
+        params = super
+        params['search'] ||= 'time >= "1 month ago"'
+    end
+
+    module PerPageAuditLimit
+
+      def self.included(base)
+        base.extend_help do |h|
+          h.section(_('Override audit record pagination')) do
+            h.text(_('Audit records for past one month are returned by default. This can be paginated further by using --per-page and --page options.'))
```
tstrachota: This is very nice example of using extended help! Thanks for that!
I think that the text needs some update though, because pagination options won't help you with listing more audits. It's passing `--search` that does the trick.

lib/hammer_cli_foreman/audit.rb
```diff
@@ -0,0 +1,76 @@
+module HammerCLIForeman
+
+  class Audit < HammerCLIForeman::Command
+
+    resource :audits
+
+    def request_params
```
tstrachota: Does it actually work? I'd expect that this method needs to be in the `Audit::ListCommand`.

lib/hammer_cli_foreman/audit.rb
rahulbajaj0509: makes sense.

lib/hammer_cli_foreman/audit.rb
rahulbajaj0509: It works as `Audit::ListComand` inherits `request_params` from `HammerCLIForeman::Comand`. So it is the same thing I guess, but it would be good to be specific that we are overriding the method for `Audit::ListCommand`.

lib/hammer_cli_foreman/audit.rb
rahulbajaj0509: ah! thanks for pointing that out :)

gnurag: How does this text look like? `Audit records for past one month are returned by default. This behavior can be changed by passing --search'`

lib/hammer_cli_foreman/audit.rb
```diff
@@ -0,0 +1,71 @@
...
+      params = super
+      params['search'] ||= 'time >= "1 month ago"'
```
ares: somehow this does not seem to apply as a default value, it still lists all audits for me

`hammer audit list --search 'time >= "1 month ago"'` when I explicitly set the search ir works, I added pry here and it was not triggered so I think @tstrachota had a valid concern above.

ares: moving to ListCommand works, but also you must add params as last line of this method, otherwise it returns the search string only

ares:
Tested, works well now, IMHO ready to be merged :+1:

rahulbajaj0509:
@mbacovsky  @ares @tstrachota thanks for your inputs throughout :)

tstrachota:
There was a small spacing issue. I didn't want to block the PR any longer so I fixed it and merged as https://github.com/theforeman/hammer-cli-foreman/commit/768148d8547a2ad9f6a190541c3ef4ad5eb913ef

This PR started on Sep 26, 2016 - wow! Thank you @gnurag and @rahulbajaj0509 for not giving up and finishing it! :champagne: I appreciate the efforts to add the functional tests.

PR #264, opened 2016-10-07
Fixes #16775 - timeout deprecation warning

tstrachota:
`nil` is required since rest-client 2.0.0 https://github.com/rest-client/rest-client/commit/4932e7e35807f98ca131412e4b74a9013b3b2dde

### DIFF
--- a/lib/hammer_cli_foreman/commands.rb
+++ b/lib/hammer_cli_foreman/commands.rb
@@ -30,7 +30,9 @@ module HammerCLIForeman
     config[:headers] = { "Accept-Language" => HammerCLI::I18n.locale }
     config[:language] = HammerCLI::I18n.locale
     config[:timeout] = HammerCLI::Settings.get(:foreman, :request_timeout)
-    config[:timeout] = -1 if (config[:timeout] && config[:timeout].to_i < 0)
+    if (config[:timeout] && config[:timeout].to_i < 0)
+      config[:timeout] = (RestClient.version < '2.0.0') ? -1 : nil
+    end
     config[:apidoc_authenticated] = false
     config
   end


xprazak2:
LGTM

lib/hammer_cli_foreman/commands.rb
```diff
@@ -30,7 +30,9 @@ def self.resource_config
     config[:headers] = { "Accept-Language" => HammerCLI::I18n.locale }
     config[:language] = HammerCLI::I18n.locale
     config[:timeout] = HammerCLI::Settings.get(:foreman, :request_timeout)
-    config[:timeout] = -1 if (config[:timeout] && config[:timeout].to_i < 0)
+    if (config[:timeout] && config[:timeout].to_i < 0)
+      config[:timeout] = (RestClient.version < '2.0.0') ? -1 : nil
```
domcleal: I think this logic would be best in apipie-bindings, would you mind moving it there? apipie-bindings has more version-dependent code and it'd also allow you to fix the deprecation warning to use `:read_timeout` instead of `:timeout`.

apipie-bindings could accept either nil or negative numbers and handle them correctly, abstracting all of these changes away from Hammer.

tstrachota:
I was actually hesitating where to put it and chose hammer in the end because the config files belog to hammer. It can equally be placed into apipie-bindings. @mbacovsky what do you think?

I created another PR https://github.com/Apipie/apipie-bindings/pull/53 and we'll choose one of them. If it's the latter hammer-cli should probably require a version of apipie-bindings that contain the patch.

mbacovsky:
I agree with domcleal. bindings already contain some version specific code and I'd prefer it to be consistent. We may turn this PR in bindings version update.

mbacovsky:
Closing this PR as it was fixed in https://github.com/Apipie/apipie-bindings/pull/53

PR #265, opened 2016-10-24
Fixes #17077 - Fix compute-resource info

ranjan:


lib/hammer_cli_foreman/compute_resource.rb
```diff
@@ -31,20 +31,19 @@ class InfoCommand < HammerCLIForeman::InfoCommand
...
+          Fields::Field.new(:label => _('UUID'), :path => [:uuid])
         ],
         'ec2' => [
-          Fields::Field.new(:label => _('Region'), :path => ["compute_resource", "region"])
+          Fields::Field.new(:label => _('Region'), :path => [:region])
         ],
         'vmware' => [
-          Fields::Field.new(:label => _('UUID'), :path => ["compute_resource", "uuid"]),
-          Fields::Field.new(:label => _('Server'), :path => ["compute_resource", "server"])
+          Fields::Field.new(:label => _('UUID'), :path => [:datacenter])
```
ares: I don't think UUID should list datacenter, UUID is identifier

lib/hammer_cli_foreman/compute_resource.rb
```diff
@@ -31,20 +31,19 @@ class InfoCommand < HammerCLIForeman::InfoCommand
 
       PROVIDER_SPECIFIC_FIELDS = {
         'ovirt' => [
-          Fields::Field.new(:label => _('UUID'), :path => ["compute_resource", "uuid"])
+          Fields::Field.new(:label => _('UUID'), :path => [:uuid])
```
ares: this does not work without also patching Foreman to include UUID in json output, please open a separate PR adding `:uuid` to https://github.com/theforeman/foreman/blob/develop/app/views/api/v2/compute_resources/ovirt.json.rabl#L1

lib/hammer_cli_foreman/compute_resource.rb
ranjan: 1. Api doesn't return attribute called uuid here is sample api response(few field truncated)
```
{"description":"","url":"vmware.example.redhat.com","name":"vmware","provider":"Vmware","provider_friendly_name":"VMware","user":"admin","datacenter":"Datacenter","server":"vmware.example.com","set_console_password":true,"images":[],"compute_attributes":[],"locations":[{"id":2,"name":"location","title":"location","description":null}],"organizations":[{"id":3,"name":"org","title":"org","description":""}]}
```
Then what should be the value of uuid?

2. Running compute resource create without uuid says "Datacenter can't be blank"
```
hammer  compute-resource create --provider="VMWare" --name "vCenter2" --description "Test vCenter"   --user "user" --password "password"  --url "vmware.example.com"
  Datacenter can't be blank
```

But after passing uuid='datacenter' allows to create vmware compute resource and works.

```
hammer compute-resource create --provider="VMWare" --name "vCenter2" --description "Test vCenter"   --user "user" --password "password"  --url "vmware.example.com" --uuid Datacenter1
Compute resource created
```
so its looks like api accept uuid as datacenter.

ares: > Api doesn't return attribute called uuid here is sample api response(few field truncated)

and the reason is, that it's missing in the API view, that's why I suggested to add `:uuid` to ovirt partial in the other comment

> But after passing uuid='datacenter' allows to create vmware compute resource and works.

I think you can enter the same identifier into either of --datacenter or --uuid, I think it's similar to --domain and --domain-id, but we don't print domain name in domain id field. If you want to display datacenter I'd suggest adding it as separate field and keeping UUID displaying UUID.

lib/hammer_cli_foreman/compute_resource.rb
ranjan: Compared code between api version v1 and v2
```
#foreman/app/views/api/v1/compute_resources/ovirt.json.rabl
attributes :user, :uuid

#foreman/app/views/api/v2/compute_resources/ovirt.json.rabl
attributes :user, :datacenter
```
```
#foreman/app/views/api/v1/compute_resources/vmware.json.rabl
attributes :user, :uuid, :server

#foreman/app/views/api/v2/compute_resources/vmware.json.rabl
attributes :user, :datacenter, :server, :set_console_password
```
It looks like they intentionally moved from uuid to datacener in v2. Any thoughts? shall i open new PR to add uuid?

ares: you're right, it's an [alias for ovirt](https://github.com/theforeman/foreman/blob/develop/app/models/compute_resources/foreman/model/ovirt.rb#L11) and [vmware](https://github.com/theforeman/foreman/blob/develop/app/models/compute_resources/foreman/model/vmware.rb#L415) so this should be most likely renamed to Datacenter in hammer for both

bottom line, the uuid on this should probably be datacenter (both label and path), other than that :+1:

ranjan:
Thanks @ares,  PR Updated

ares:
Works for me now, thanks @ranjan, :+1: from me.

tstrachota:
Tested the PR and it works fine. Thanks @ranjan!

PR #267, opened 2016-11-09
The 'start' key always needs a 1 or 0

wzzrd:
Previously, the docs mentioned that the 'start' key for compute-attributes is
a boolean. This suggests values like 'true', 'false', 'yes' and 'no' would be
accepted.

This was not the case. For libvirt (which is what I tested with, I assume oVirt
is the same), a 'true' or 'false' value did nothing, only a 1 or 0 did.

This patch fixes the docs to explain a 0 or 1 is required to influence
autostart behaviour.

### DIFF
--- a/doc/host_create.md
+++ b/doc/host_create.md
@@ -233,7 +233,7 @@ path                 # path to folder
 guest_id             # guest OS id form VMware
 scsi_controller_type # id of the controller from VMware
 hardware_version     # hardware version id from VMware
-start                # Must be a 1 or 0, whether to start the machine or not
+start                # Boolean, expressed as 0 or 1, whether to start the machine or not
 ```
 
 Available keys for `--interface`:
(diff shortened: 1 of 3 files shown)


tstrachota:
Thanks @wzzrd We've recently merged improved help for host create. Would you mind fixing it there too?
* libvirt: https://github.com/theforeman/hammer-cli-foreman/blob/master/lib/hammer_cli_foreman/compute_resources/libvirt/host_help_extenstion.rb#L14
* oVirt: https://github.com/theforeman/hammer-cli-foreman/blob/master/lib/hammer_cli_foreman/compute_resources/ovirt/host_help_extenstion.rb#L16

wzzrd:
@tstrachota, I updated those two locations as well, and made documentation consistent over all three locations.

tstrachota:
I addressed one more docs-related copy-paste typo and merged as: https://github.com/theforeman/hammer-cli-foreman/commit/3a8c32b44515fcde15abe4a9a70366b3cf95d08b
Thanks @wzzrd !

PR #268, opened 2016-11-16
Fixes #17355 - display override value order as long text

tstrachota:


ares:
:+1: from me, works nicely

this 
```
    Order:               fqdn, hostgroup, os, domain
```
changed to this
```
    Order:               
      fqdn
      hostgroup
      os
      domain
```

PR #269, opened 2016-11-24
Fixes #8016 - Session auth in hammer

tstrachota:
- api connection moved to a separate object
- credentials moved to InteractiveBasicAuth
- new SessionAuthenticatorWrapper for handling api sessions

Requires:
- https://github.com/Apipie/apipie-bindings/pull/52
- https://github.com/theforeman/hammer-cli/pull/227
- https://github.com/theforeman/foreman/pull/4045

TODO: 
- [x] add logging of auth type
- [x] add config for turning sessions off
- [x] fix sessions being dropped after unauthenticated call (`hammer ping`)


mbacovsky:
First of all thanks for this cool feature, it was on many wish lists for a long time.

I was testing it and it looks almost ready to go. I've found the following that needs some solution/discussion:

1. it would be great to know from hammer logs what auth was used - session, credentials
1. can we influence the expiration time from hammer side?
1. can we turn the session auth feature off?
1. can we set the session expired from hammer side?
1. when session expires hammer fails with error. It should retry and obtain credentials from config or user as usually

As per our discussion I know that some might be difficult to implement right now. Filed issue to track the change may be acceptable solution in some cases. 

What is really annoying is that session expires when using command that does not need auth (e.g. ping) and it makes the following command fail. Can that be fixed?

tstrachota:
Thanks for the review, answers below:

_1. it would be great to know from hammer logs what auth was used - session, credentials_
Ok, I'll add some more logging.

_2. can we influence the expiration time from hammer side?_
No

_3. Can we turn the session auth feature off?_
No, it probably makes sense to add it.

_4. can we set the session expired from hammer side?_
No

_5. when session expires hammer fails with error. It should retry and obtain credentials from config or user as usually_
I agree such behaviour would be much better. I intentionally didn't introduce it in this PR because that would make it even more complicated and difficult to review. It's something that's definitely worth adding in future.

Expiring the session after unauthenticated call is definitely a bug. I'll fix that. Please feel free to open additional issues for the rest.

mbacovsky:
@tstrachota, I did some more testing. The session expiration is set by default to 60 min. It is very annoying When after every hour of inactivity your hammer command fail. I also used hammer in a cron job and every second run failed when the interval was bigger then the expiration time. We either need to implement the retry or add possibility to disable sessions by configuration and CLI arg. What do you think?

tstrachota:
@mbacovsky updated and ready for another round of review
* issues 1. and 3. from https://github.com/theforeman/hammer-cli-foreman/pull/269#issuecomment-264854529 fixed
* fixed logout after `hammer ping`
* `hammer auth *` commands fixed to work with new authenticators
* authenticator now handles change of user when there's existing session

Sessions are now turned off by default and you have to enable it by setting `:use_sessions: true` in the config file.

config/foreman.yml
```diff
@@ -8,6 +8,7 @@
   # Credentials. You'll be asked for them interactively if you leave them blank here
   :username: 'admin'
   #:password: 'example'
+  :use_sessions: false
```
mbacovsky: Are sessions specific to Basic Auth or can they be used with e.g. OAuth? Just to confirm this is the right place for the config.

mbacovsky:
I've tested this and now it works well for me. Thanks for fixing the auth commands, I forgot to test them.
I'm ready to merge when the question on the config is answered.

config/foreman.yml
tstrachota: Some auth methods might require sessions but in general it's imho independent.

tstrachota:
@mbacovsky thanks. I'll bump dependency on the latest apipie-bindings prior to merge.

mbacovsky:
[test]

mbacovsky:
@tstrachota this looks good to merge. Please add deps on apipie-bindings >= 0.0.19 that was released today. Part in core was merged as it seems it can work with older bindings.

tstrachota:
@mbacovsky thanks for the apipie-bindings release! Dependency bumped.

mbacovsky:
@tstrachota awesome!
lgtm

thomasmckay:
@tstrachota @mbacovsky - I need some help here. How do I replace this file with the new mechanics for auth?
https://github.com/Katello/hammer-cli-csv/blob/master/lib/hammer_cli_csv/utils/config.rb
http://projects.theforeman.org/issues/17624

PR #270, opened 2016-12-03
Fixes #17568 - create compute-resource VMware - URL optional

swapab:


mbacovsky:
@swapab, thanks for the contribution. The code looks good, but I've tested this and I'm getting the following from the server:
```
$ hammer compute-resource create --name xxx --provider vmware
  URL can't be blank
  Username can't be blank
  Password can't be blank
  Server can't be blank
  Datacenter can't be blank
```
It seems both  `server` [1] and `url` [2] is required on the server. So this patch indeed does not fix the issue. I think it has to be sorted out on the server first and then aligned in hammer. To keep the original behavior at least the `server` should be required by hammer for vmware (preferably the whole required set).

[1] https://github.com/theforeman/foreman/blob/develop/app/models/compute_resources/foreman/model/vmware.rb#L8
[2] https://github.com/theforeman/foreman/blob/develop/app/models/compute_resource.rb#L18

swapab:
Due to bad merge, I had to delete the branch.
The new pull request is https://github.com/theforeman/hammer-cli-foreman/pull/272

ohadlevy:
you can always force push into an exiting remote branch :)

swapab:
Thanks @ohadlevy 
I did that actually. But somehow I messed up the commits order and ended up having 2 commits with same changes entangled with commits from master's.

ldomb:
This also affects the ec2 provider as well. Hammer expect a url when adding an ec2 provider

PR #272, opened 2016-12-15
Fixes #17568 - create compute-resource VMware - URL optional

swapab:


mbacovsky:
@swapab, thanks for the contribution. The code looks good, but I've tested this and I'm getting the following from the server:
```
$ hammer compute-resource create --name xxx --provider vmware
  URL can't be blank
  Username can't be blank
  Password can't be blank
  Server can't be blank
  Datacenter can't be blank
```
It seems both  `server` [1] and `url` [2] is required on the server. So this patch indeed does not fix the issue. I think it has to be sorted out on the server first and then aligned in hammer. To keep the original behavior at least the `server` should be required by hammer for vmware (preferably the whole required set).

[1] https://github.com/theforeman/foreman/blob/develop/app/models/compute_resources/foreman/model/vmware.rb#L8
[2] https://github.com/theforeman/foreman/blob/develop/app/models/compute_resource.rb#L18

lib/hammer_cli_foreman/compute_resource.rb
```diff
@@ -70,17 +70,29 @@ def print_data(data)
 
     class CreateCommand < HammerCLIForeman::CreateCommand
 
+      VALIDATION_PER_PROIVIDER = {
+        'libvirt'   => [:option_url],
+        'ovirt'     => [:option_url, :option_user, :option_password, :option_datacenter],
+        'openstack' => [:option_url, :option_user, :option_password],
+        'racksapce' => [:option_url],
+        'ec2'       => [:option_region, :option_user, :option_password],
+        'vmware'    => [:option_user, :option_password, :option_datacenter, :option_server]
```
swapab: @mbacovsky Thanks for reviewing my PR.

This is kind of vague but working solution to validate ComputeResource#create options. One thing to note in case of vmware[1] and ec2 URL option has been *smartly* bypassed with server and zone respectively.

[1] https://github.com/theforeman/foreman/blob/develop/app/models/compute_resources/foreman/model/vmware.rb#L423

I'll create a new patch to handle server side url validation(url should be validated based on provider)

tstrachota:
Hi @swapab, has there been any progress with the server-side patch, please?

swapab:
@tstrachota Server side patch is ready: https://github.com/theforeman/foreman/pull/4461

swapab:
@mbacovsky @tstrachota Since the pre-requisite theforeman/foreman#4461 has been merged. This PR is ready for review.

lib/hammer_cli_foreman/compute_resource.rb
```diff
@@ -70,17 +70,29 @@ def print_data(data)
...
       failure_message _("Could not create the compute resource")
 
       build_options
 
       validate_options do
-        all(:option_name, :option_url, :option_provider).required
+        all(
+          *VALIDATION_PER_PROIVIDER[
+            option(:option_provider).value.to_s.downcase
+          ].unshift(:option_name, :option_provider)
```
tstrachota: I get an error when I try to create a compute resource without specifying the provider:
```
hammer compute-resource create --name Test
  Error: undefined method `unshift' for nil:NilClass
```
the same happens when I use some unexpected value for the provider, eg. `--provider=unknown`

swapab: @tstrachota Thanks for reviewing. I have fixed the above error. Functional test added.

.gitignore
```diff
@@ -22,3 +22,4 @@ tags
 # Locale files
 locale/*/*.edit.po
 locale/*/*.po.time_stamp
+.tramp_history
```
swapab: @tstrachota Let me know if this can't be added to `.gitignore`

tstrachota: This is totally fine. Just move it below the comment "#Editors" so that people roughly know what it belongs to.

tstrachota:
@swapab thank you for the updates, especially for the tests. They look good! Could you please add two more that would specifically test the provider option being `--provider=""` and `--provider="unknown"`. The latter will imho still fail with `"undefined method `unshift' for nil:NilClass"`.

swapab:
@tstrachota 
Thanks again. I have updated the PR.
1. Moved `.tamp_history` to under #Editor 
2. Fixed erro for `invalid/blank` provider
---
Tests are failing though http://ci.theforeman.org/job/test_hammer_cli_foreman_pull_request/1056/
```
user::create.test_0001_resolves default organization name to id
user::create.test_0002_resolves default location name to id
```

tstrachota:
@swapab thanks for the changes. The test failures are unrelated, the master is broken since the merge of https://github.com/theforeman/hammer-cli/commit/62529bb8b918e0bae6cda7592c1a9c0afce6e8b4 We'll deal with that in a separate PR.

mbacovsky:
@tstrachota, @swapab , the PR fixing the tests is https://github.com/theforeman/hammer-cli-foreman/pull/333

tstrachota:
[tests]

tstrachota:
Merged, thanks @swapab ! I'm sorry for the delay between approving the PR and merging it.

swapab:
Yay! Thanks @tstrachota

PR #274, opened 2016-12-19
Fixes #17730 - prevent from setting taxonomies for non-overriding filters

tstrachota:
* signals usage error when org/loc options are used and --override is
  false
* resets taxonomies when filter's override flag is modified to false
* extends help with usage description

### DIFF
--- a/lib/hammer_cli_foreman/filter.rb
+++ b/lib/hammer_cli_foreman/filter.rb
@@ -41,17 +41,78 @@ module HammerCLIForeman
       build_options
     end
 
+
+    module TaxonomyCheck
+      def taxonomy_options?
+        option_location_names || option_location_ids || option_organization_names || option_organization_ids
+      end
+
+      def signal_override_usage_error
+        signal_usage_error _('Organizations and locations can be set only for overriding filters')
+      end
+
+      def self.included(base)
+        base.extend_help do |h|
+          h.section('Overriding organizations and locations') do
+            override_condition = "--override=true"
+            org_opts = '--organization[s|-ids]'
+            loc_opts = '--location[s|-ids]'
+
+            h.text(_("Filters inherrit organizations and locations from it's role by default. This behavior can be changed by setting %{condition}\n"+
+              "Therefore options %{org_opts} and %{loc_opts} are applicable only when the override flag is set.") % {
+              :org_opts => org_opts,
+              :loc_opts => loc_opts,
+              :condition => override_condition
+            })
+          end
+        end
+      end
+    end
+
+
     class CreateCommand < HammerCLIForeman::CreateCommand
+      include TaxonomyCheck
+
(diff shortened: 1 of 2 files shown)


lib/hammer_cli_foreman/filter.rb
```diff
@@ -41,18 +41,79 @@ def extend_data(filter)
...
+      end
+
+      def self.included(base)
+        base.extend_help do |h|
+          h.section('Overriding organizations and locations') do
+            override_condition = "--override=true"
+            org_opts = '--organization[s|-ids]'
+            loc_opts = '--location[s|-ids]'
+
+            h.text(_("Filters inherrit organizations and locations from it's role by default. This behavior can be changed by setting %{condition}\n"+
```
ares: s/inherrit/inherit/

lib/hammer_cli_foreman/filter.rb
```diff
@@ -41,18 +41,79 @@ def extend_data(filter)
...
+        if override?
+          # Clear taxonomies in case the filter is switching override from true to false
+          params['filter']['location_ids'] = []
+          params['filter']['organization_ids'] = []
+        end
+        params
+      end
+
+      def execute
+        signal_override_usage_error if !override? && taxonomy_options?
```
ares: this seems to be defined in https://github.com/theforeman/hammer-cli-foreman/pull/274/files#diff-fda455aadf445eef31212357cc20c85cR79 as well, it's included already?

tstrachota: Not sure I understand. It's defined in a different class and unfortunately they don't inherit from each other. Or are you asking for moving it to the `TaxonomyCheck` module and aliasing the method chain?

lib/hammer_cli_foreman/filter.rb
```diff
@@ -41,18 +41,79 @@ def extend_data(filter)
...
+      end
+
+      def self.included(base)
+        base.extend_help do |h|
+          h.section('Overriding organizations and locations') do
+            override_condition = "--override=true"
+            org_opts = '--organization[s|-ids]'
+            loc_opts = '--location[s|-ids]'
+
+            h.text(_("Filters inherit organizations and locations from it's role by default. This behavior can be changed by setting %{condition}\n"+
```
ares: I think it should be `its` instead of `it's`.

ares:
Except one nit I think it's ready for merge, I tested it and works well.

tstrachota:
Tests were failing because they are executed against old version of apidoc. Unfortunately migration to the latest apidoc isn't smooth and requires some additional changes in other tests too. I've added a temporary fix for this PR and I'll deal with migration to apidoc 1.14 in a separate PR.

ares:
:+1: for commits above

PR #279, opened 2017-01-22
Fixes #6344 - Add compute profile command

shlomizadok:
@tstrachota -  would love some help with this (see questions in code)

### DIFF
--- a/lib/hammer_cli_foreman/compute_profile.rb
+++ b/lib/hammer_cli_foreman/compute_profile.rb
@@ -1,0 +1,69 @@
+require 'hammer_cli_foreman/image'
+
+module HammerCLIForeman
+
+  class ComputeProfile < HammerCLIForeman::Command
+    resource :compute_profiles
+
+    class ListCommand < HammerCLIForeman::ListCommand
+
+      output do
+        field :id, _("Id")
+        field :name, _("Name")
+      end
+
+      build_options
+    end
+
+    class InfoCommand < HammerCLIForeman::InfoCommand
+
+      output ListCommand.output_definition do
+
+        HammerCLIForeman::References.taxonomies(self)
+        HammerCLIForeman::References.timestamps(self)
+
+        collection :compute_attributes, _("Compute attributes") do
+          field :id, _('Id')
+          field :name, _('Name')
+          field nil, _("Compute Resource"), Fields::SingleReference, :key => :compute_resource
+          field :vm_attrs, _("VM attributes")
+        end
+      end
+
+      build_options
+    end
+
+
+    class CreateCommand < HammerCLIForeman::CreateCommand
+      success_message _("Compute resource created")
+      failure_message _("Could not create the compute resource")
+
+      build_options
+
+      validate_options do
+        all(:option_name).required
+      end
+    end
+
+
+    class UpdateCommand < HammerCLIForeman::UpdateCommand
+      success_message _("Compute resource updated")
+      failure_message _("Could not update the compute resource")
+
+      build_options :without => :name
+    end
+
+
+    class DeleteCommand < HammerCLIForeman::DeleteCommand
+      success_message _("Compute resource deleted")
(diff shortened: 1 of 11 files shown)


lib/hammer_cli_foreman/compute_profile.rb
```diff
@@ -0,0 +1,80 @@
...
+    end
+
+
+    class InfoCommand < HammerCLIForeman::InfoCommand
+
+      output ListCommand.output_definition do
+        HammerCLIForeman::References.taxonomies(self)
+        HammerCLIForeman::References.timestamps(self)
+
+        collection :compute_attributes, _("Compute attributes") do
```
shlomizadok: Can I have non-numbered collection?

lib/hammer_cli_foreman/compute_profile.rb
```diff
@@ -0,0 +1,80 @@
...
+          field :id, _('Id')
+          field :name, _('Name')
+          field :compute_resource_id, _('Compute Resource Id')
+          field :compute_resource_name, _('Compute resource name')
```
shlomizadok: API also returns `compute_profile_{id,name}` do you think I should add them here too? Or `Name`, `Id` are sufficient?

lib/hammer_cli_foreman/compute_profile.rb
tstrachota: Yes, use:
```ruby
collection :compute_attributes, _("Compute attributes"), :numbered => false do
```

lib/hammer_cli_foreman/compute_profile.rb
tstrachota: I would omit `compute_profile_{id,name}` because the value will be the same for all items in the collection.

BTW the standard way of displaying names and ids of referenced items in hammer is:
```ruby
field nil, _("Compute Resource"), Fields::SingleReference, :key => :compute_resource
```

lib/hammer_cli_foreman/compute_profile.rb
```diff
@@ -0,0 +1,69 @@
...
+      output ListCommand.output_definition do
+
+        HammerCLIForeman::References.taxonomies(self)
+        HammerCLIForeman::References.timestamps(self)
+
+        collection :compute_attributes, _("Compute attributes") do
+          field :id, _('Id')
+          field :name, _('Name')
+          field nil, _("Compute Resource"), Fields::SingleReference, :key => :compute_resource
+          field :vm_attrs, _("VM attributes")
```
shlomizadok: @tstrachota - tried for like a day to normalize (automatically) this collection.
We could leave this as it is, or if you have an idea - I'd love to hear.

shlomizadok:
@tstrachota, currently `vm_attrs` would return as a long json (which is an ok- solution).
I have opened a PR on core (https://github.com/theforeman/foreman/pull/4231) that will help normalize the `vm_attrs` data.

We can merge this now and add a new PR after ^ is merged - or wait for this one

lib/hammer_cli_foreman/compute_profile.rb
tstrachota: @shlomizadok I have an idea. Patch is on the way...

tstrachota: @shlomizadok see https://github.com/shlomizadok/hammer-cli-foreman/pull/1
The code will need more care. I'm sending it to share the idea rather than presenting a final solution.

tstrachota:
My patch should take care of formatting even when there's no `provider_friendly_name` in the server response.

tstrachota:
Since my patches were merged into this PR it now depends on :
https://github.com/theforeman/hammer-cli/pull/232

tstrachota:
This PR is stalled so I'm going to close it. I took the feature over from @shlomizadok, but the progress is slow. I'll open a new PR once it's ready for review.

PR #281, opened 2017-02-01
Fixes #18337 - skip generating option --root-pass for host create

tstrachota:


ares:
ACK, only suggestion I have is to either open refactoring issue to remove this deprecated option later or document it in code so we remove it one day. But it depends on how you do that normally in hammer... so :+1: from me

tstrachota:
@ares thanks, FYI I opened: http://projects.theforeman.org/issues/18636

PR #282, opened 2017-02-09
fixes #18439 - only include .mo files below locale/

domcleal:


tstrachota:
Nice, the gem is significantly smaller with this patch. Thanks @domcleal !

PR #283, opened 2017-02-09
Fixes #18318 - Add command to clone user role

xprazak2:
Needs [new endpoint](https://github.com/theforeman/foreman/pull/4267) to work


tstrachota:
```
hammer role clone -h
Usage:
    hammer role clone [OPTIONS]

Options:
 --id ID                        
 --name NAME                   User role name
 -h, --help                    print help
```
There's no option for passing the new name (`--name` is be used as current role's identifier). Inheriting from `HammerCLIForeman::UpdateCommand` should add option `--new-name` and make the command working as expected.

tstrachota:
BTW it's a different story, but when one tries to execute the command against foreman without the required patch it correctly fails with:
```
Could not clone the user role:
  Error: The server does not support such operation.
```
But help isn't very useful in this case:
```
hammer role clone -h
Usage:
    hammer role clone [OPTIONS]

Options:
 -h, --help                    print help
```
I think we should rather display something like:

```
hammer role clone -h
Usage:
    hammer role clone [OPTIONS]

Unfortunately the server does not support such operation.
```
I'll create a separate issue for that. It should be fixed in the core anyway.

tstrachota:
@xprazak2 could you please add some simple functional tests for the command?

tstrachota:
FYI http://projects.theforeman.org/issues/18532

xprazak2:
`CloneCommand` inherits from `UpdateCommand` now and has the `--new-name` option, I also added functional tests.

test/functional/role_test.rb
```diff
@@ -0,0 +1,72 @@
...
+      params = ['--name=old', '--new-name=zzz']
+
+      api_expects(:roles, :index, 'Attempt find role') do |par|
+        par[:search] == 'name = "old"'
+      end.returns(
+        index_response(
+          [{
+              'name' => 'old',
+              'id' => 1
+          }]))
```
tstrachota: Not a big deal but note that you can also use
```ruby
api_expects_search(:roles, { :name => 'old' }).returns(...)
```
here.

tstrachota: Or
```ruby
api_expects_search(:roles, { :name => 'old' },  'Attempt find role').returns(...)
```
if you want to have the message there.

test/functional/role_test.rb
```diff
@@ -0,0 +1,72 @@
...
+      result = run_cmd(@cmd + params)
+
+      assert_cmd(expected_result, result)
+    end
+
+    it 'should clone a role by id' do
+      params = ['--id=1', '--new-name=zzz']
+
+      api_expects(:roles, :clone, 'Clone role') do |par|
+        par['id'] == '1' && par['new_name'] == 'zzz'
```
tstrachota: I noticed that the apidoc exported for the tests is different from what's currently in https://github.com/theforeman/foreman/pull/4267 The server side parameter is just `name`, not `new_name`.

From the tests:
```
[3] pry(#<HammerCLIForeman::Role::CloneCommand>)> resource.action(:clone).params
=> [<Param *id (String)>, <Param *new_name (String)>]
```

We should update the api export and test `par['name'] == 'zzz'` here.

tstrachota:
@xprazak2 thanks for the changes, it looks good now. Only the tests need updated api export.

test/functional/role_test.rb
xprazak2: I think we are good here, the api export is already updated, I just did not push the new version of api to [theforeman/foreman#4267](https://github.com/theforeman/foreman/pull/4267) for some reason - fixed now. The test works as expected, role should be cloned when `--id` and `--new-name` are supplied, where `new-name` is a [name of the role that is being created](https://github.com/theforeman/hammer-cli-foreman/pull/283#issuecomment-280292557).

ares: the API PR uses `name`, added a comment there

tstrachota: I see, thanks for explanation @xprazak2. LGTM now but I'll hold on with merging this PR until the API counterpart is merged. Just to be sure there are no more changes.

ares: API PR has been merged

test/functional/role_test.rb
```diff
@@ -0,0 +1,72 @@
...
+      result = run_cmd(@cmd + params)
+      assert_cmd(success_result("User role cloned\n"), result)
+    end
+
+    it 'should clone a role by name' do
+      params = ['--name=old', '--new-name=zzz']
+
+      api_expects_search(:roles, { :name => 'old' }, 'Attempt find role') do |par|
+        par[:search] == 'name = "old"'
+      end.returns(
```
tstrachota: The block is redundant here. `api_expects_search` sets the same expectation, even with better error reporting.

test/functional/role_test.rb
tstrachota: I do think this expectation is wrong. Tests fail for me with apidoc export from latest master (with the api counterpart merged). Please make sure the PR contains latest apidoc export.
```
  1) Failure:
role::clone#test_0003_should clone a role by id [/usr/local/rvm/gems/ruby-2.3.0/gems/apipie-bindings-0.0.19/lib/apipie_bindings/api.rb:174]:
unexpected invocation: #<AnyInstance:ApipieBindings::API>.call_action(<Action roles:clone>, {'id' => '1', 'role' => {'name' => 'zzz'}, 'name' => 'zzz'}, {}, {:fake_response => nil})
unsatisfied expectations:
- expected exactly once, not yet invoked: Clone role
  #<AnyInstance:ApipieBindings::API>.call_action(:roles, :clone, &block)

  2) Failure:
role::clone#test_0004_should clone a role by name [/usr/local/rvm/gems/ruby-2.3.0/gems/apipie-bindings-0.0.19/lib/apipie_bindings/api.rb:174]:
unexpected invocation: #<AnyInstance:ApipieBindings::API>.call_action(<Action roles:clone>, {'id' => 1, 'role' => {'name' => 'zzz'}, 'name' => 'zzz'}, {}, {:fake_response => nil})
unsatisfied expectations:
- expected exactly once, not yet invoked: Clone role
  #<AnyInstance:ApipieBindings::API>.call_action(:roles, :clone, &block)
satisfied expectations:
- expected exactly once, invoked once: Attempt find role
  #<AnyInstance:ApipieBindings::API>.call_action(:roles, :index, &block)
```

One more note, not important one, just FYI:
The block param matchers can now be replaced with `.with_params`. The advantage is getting much better error report (diff of params) when the expectation fails. See [testing api expectations docs](https://github.com/theforeman/hammer-cli-foreman/blob/master/doc/testing.md#available-helper-functions) for more details.

lib/hammer_cli_foreman/role.rb
```diff
@@ -71,6 +71,23 @@ class UpdateCommand < HammerCLIForeman::UpdateCommand
       build_options
     end
 
+    class CloneCommand < HammerCLIForeman::UpdateCommand
+      action :clone
+
+      def self.command_name(name = "clone")
```
tstrachota: In this case, it can be shortened to just `command_name "clone"`. But this works too plus it can be overridden from subclasses so NP.

xprazak2:
I made changes on foreman side and this got outdated again 😢. I updated the api export, used `with_params` method instead of blocks and hopefully tests will turn green.

tstrachota:
:+1: looks good now, thanks @xprazak2 !

PR #285, opened 2017-02-21
Fixes #17923 - user create/update accepts org/loc name

tstrachota:


ares:
ACK, works nicely :+1:

PR #287, opened 2017-03-01
Fixes #18739 - Display hostgroup_title on Host actions

dLobatog:
Before this patch, host list was displaying only the leaf of the
hostgroup tree. E.g:

base/hostgroup1

displayed as 'hostgroup1' in the Host list table.

After this patch, 'base/hostgroup1' is displayed, which is more useful
than the leaf (which could be the same for different trees). The same
thing has been fixed for Host info.

@tstrachota I also wonder if we should do this for organizations/locations

### DIFF
--- a/lib/hammer_cli_foreman/host.rb
+++ b/lib/hammer_cli_foreman/host.rb
@@ -30,7 +30,7 @@ module HammerCLIForeman
         field :id, _("Id")
         field :name, _("Name")
         field nil, _("Operating System"), Fields::SingleReference, :key => :operatingsystem
-        field :hostgroup_title, _("Host Group")
+        field nil, _("Host Group"), Fields::SingleReference, :key => :hostgroup, :display_field => 'title'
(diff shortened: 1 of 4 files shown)


ares:
Works well and fixes the issue for me :+1: I agree it would be good to do it the same for orgs/locs, the structure is not clear in their org/loc list

tstrachota:
This is nice improvement. I'm afraid it breaks `--show-ids` functionality for hostgroup references though. @dLobatog how about something like this: https://github.com/dLobatog/hammer-cli-foreman/pull/3

dLobatog:
@tstrachota That's what I tried initially, I was missing this part, thanks for that 😄  (https://github.com/dLobatog/hammer-cli-foreman/pull/3/files#diff-99b175729e323d7a7b894d35995e0c60R13). The PR is updated (credits due in your commit)

tstrachota:
:+1:

PR #289, opened 2017-03-06
Fixes #18005 - Fix Hammer-cli current user password update

dhlavac:


ares:
[test]

lib/hammer_cli_foreman/user.rb
```diff
@@ -89,6 +89,25 @@ class UpdateCommand < HammerCLIForeman::UpdateCommand
       failure_message _("Could not update the user")
 
       include CommonUpdateOptions
+
+      option "--current-password", "CURRENT_PW", " "
+      option "--ask-current-password", "ASK_CURRENT_PW", " ",:format => HammerCLI::Options::Normalizers::Bool.new
+
+      def self.ask_current_password
+        prompt = _("Enter User's current password :") + " "
```
tstrachota: Lower 'u' in "User's" and remove space in front of the colon please.

tstrachota:
I played with the command and found some usability issues that are worth improving:
1. We should enable asking for a new password interactively too. I propose to replace `--ask-current-password` with `--ask-password` and use it for both values:
    * `!option_password && option_ask_password` => ask for new password interactively
       (similar behaviour should apply to `user create` btw)
    * `!option_current_password && option_ask_password` => ask for the current password interactively
2. When a user without admin permissions attempts to change password without entering the current one, server replies with error "Current password Incorrect password". I think it's not that obvious what a user should do in such case.

I propose to detect whether the current user (the one that authenticates from hammer) has admin permissions and check whether current password is present if the user isn't an admin.
Something like:
```ruby
      def current_user
        HammerCLIForeman.foreman_api_connection.resource(:users).call(:show, :id => HammerCLIForeman.foreman_api_connection.authenticator.user(true))
      end

      def validate_options
        super
        if !options['option_ask_password'] && options['option_password'] && !current_user['admin']
          validator.option(:option_current_password).required
        end
      end
```
This would require change in `InteractiveBasicAuth` authenticator to ask for a user when `true` is passed to its `#user` method.

tstrachota:
Note for other potential reviewers:
This PR requires https://github.com/theforeman/foreman/pull/4359 (+ running `rake apipie:cache`) to work properly.

dhlavac:
@tstrachota changes done, can you look at that

lib/hammer_cli_foreman/user.rb
```diff
@@ -89,6 +105,20 @@ class UpdateCommand < HammerCLIForeman::UpdateCommand
       failure_message _("Could not update the user")
 
       include CommonUpdateOptions
+
+      def request_params
+        params = super
+        require 'pry'
+        #binding.pry
```
tstrachota: Some leftovers after debugging

dhlavac:
@tstrachota removed debug code

tstrachota:
[test]

dhlavac:
@tstrachota It should works now, can you please look at that and run test ?

ares:
[test]

ares:
tests are green, what do you think @tstrachota ?

tstrachota:
Btw for users without a role `Viewer` the workflow fails on "User not found". This is caused by http://projects.theforeman.org/issues/20112 The issue can be fixed separately. No need to block this PR.

dhlavac:
@tstrachota comments resolved and added some more tests. Please look for that if test coverage is enough

ares:
add to PR whitelist

ares:
[test]

lib/hammer_cli_foreman/api/session_authenticator_wrapper.rb
```diff
@@ -62,7 +62,7 @@ def response(r)
         @authenticator.response(r)
       end
 
-      def user
+      def user(ask=nil)
         @authenticator.user if @authenticator.respond_to?(:user)
```
tstrachota: Please pass `ask` to the encapsulated authenticator

tstrachota:
@dhlavac one more note and then I'd say we can merge

xprazak2:
[test]

tstrachota:
[test]

tstrachota:
:+1: this is ready to merge when the tests are green (the failure is unrelated).
Good job @dhlavac !

PR #290, opened 2017-03-13
Fixes #18878 - Adding hidden_value option to parameters

orrabin:


orrabin:
This works for global parameters but it doesn't hide host parameters yet.

lib/hammer_cli_foreman/common_parameter.rb
```diff
@@ -26,6 +26,7 @@ class SetCommand < HammerCLIForeman::Command
 
       option "--name", "NAME", _("parameter name"), :required => true
       option "--value", "VALUE", _("parameter value"), :required => true
+      option "--hidden_value", "HIDDEN VALUE", _("should the value be hidden")
```
mbacovsky: Options should have dashes only:  `--hidden-value`. Multi-word param names should be connected with underscore: ` HIDDEN_VALUE`.

lib/hammer_cli_foreman/parameter.rb
```diff
@@ -48,6 +48,7 @@ def self.create_option_builder
     class SetCommand < AbstractParameterCommand
       option "--name", "NAME", _("parameter name"), :required => true
       option "--value", "VALUE", _("parameter value"), :required => true
+      option "--hidden_value", "HIDDEN VALUE", _("should the value be hidden")
```
mbacovsky: dtto (dash and underscore missing)

orrabin:
@mbacovsky thanks, fixed

lib/hammer_cli_foreman/common_parameter.rb
```diff
@@ -26,6 +26,7 @@ class SetCommand < HammerCLIForeman::Command
 
       option "--name", "NAME", _("parameter name"), :required => true
       option "--value", "VALUE", _("parameter value"), :required => true
+      option "--hidden-value", "HIDDEN_VALUE", _("should the value be hidden")
```
mbacovsky: While testing I found out one usability issue - the new option should be of type boolean so that the the allowed values are shown in the help and the value is validated.
```ruby
      option "--hidden-value", "HIDDEN_VALUE", _("should the value be hidden"),
        :format => HammerCLI::Options::Normalizers::Bool.new
```

lib/hammer_cli_foreman/parameter.rb
```diff
@@ -48,6 +48,7 @@ def self.create_option_builder
     class SetCommand < AbstractParameterCommand
       option "--name", "NAME", _("parameter name"), :required => true
       option "--value", "VALUE", _("parameter value"), :required => true
+      option "--hidden-value", "HIDDEN_VALUE", _("should the value be hidden")
```
mbacovsky: dtto - boolean

mbacovsky:
@orrabin, Thanks for updating. I requested one more change that I found out during my testing.

orrabin:
@mbacovsky I added tests but they are not working yet. Adding `hidden-value` to the api mock resulted an empty string instead of an empty hash sent to `print_message` which results in the error: `TypeError: no implicit conversion of Symbol into Integer`.

mbacovsky:
:+1: Thanks @orrabin, it looks and works great now! Merging... :bouquet:

PR #295, opened 2017-04-03
fixes #19134 - add vlan id field to subnet

timogoebel:


ares:
:+1: from me, works fine

tstrachota:
Thanks @timogoebel 
We already have that field defined for the `info` command, which inherits the field definition from `list` (https://github.com/timogoebel/hammer-cli-foreman/blob/7ec4ca90252ec3babb87ee70520e440af0914d8c/lib/hammer_cli_foreman/subnet.rb#L31). Can you please remove vlan id from `info` in your commit to prevent it from being printed twice? We should probably also keep the label in capitals for backward compatibility.

timogoebel:
@tstrachota : Updated. Thanks. Did not spot that...

ares:
oh I missed that but I can confirm that it now displays just once

tstrachota:
Thank you @timogoebel

PR #296, opened 2017-04-03
Refs #19083 - Install server CA cert without root access

mbacovsky:
This patch updates the SSL verification error message according to the
related changes in Hammer core

### DIFF
--- a/lib/hammer_cli_foreman/exception_handler.rb
+++ b/lib/hammer_cli_foreman/exception_handler.rb
@@ -96,25 +96,11 @@ module HammerCLIForeman
 
       cmd = "hammer --fetch-ca-cert #{host}"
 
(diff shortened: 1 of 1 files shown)


mbacovsky:
Complements the core PR https://github.com/theforeman/hammer-cli/pull/236

mbacovsky:
Updated error message:
```
hammer -s https://my.server.com user list
SSL certificate verification failed
Make sure you configured the correct URL and have the server's CA certificate installed on your system.

You can use hammer to fetch the CA certificate from the server. Be aware that hammer cannot verify whether the certificate is correct and you should verify its authenticity after downloading it.

Download the certificate as follows:

  $ hammer --fetch-ca-cert https://my.server.com

```

tstrachota:
:+1: seems fine to me. Should be merged together with its counterpart in `hammer-cli`

PR #297, opened 2017-04-04
fixes #19156 - compute-resources has images subcommand

timogoebel:


tstrachota:
Nice addition @timogoebel 
Would you mind adding a check if `--name` and `--id` options are present?
```ruby
validate_options do
  any(:option_id, :option_name).required
end
```

FYI we started adding tests for new commands into `functional/`. They're a bit more verbose but enable testing in the context of whole cli and provide more readable failure messages when things go wrong.

timogoebel:
@tstrachota : Thanks. Added the validations.

tstrachota:
@timogoebel cool. The tests failed because they now expect the options.

timogoebel:
@tstrachota : Fixed the tests, thanks.

tstrachota:
Merged, thanks @timogoebel !

PR #299, opened 2017-04-04
Fixes #19157 - Enable taxonomy titles

tstrachota:
Names of locations and organizations aren't globally unique. Without
titles it's impossible to reference organizations or locations with the
same name nested in different parent.

Changes:
- adds Title to output of organization and location list
- adds `--location-title[s]` and `--organization-title[s]` to all commands that
  consume taxonomies
- print titles instead of names for taxonomy references


tstrachota:
An alternative approach is to completely replace names with labels in hammer internally. We would have to change id resolver and let the list commands print titles (in the labeled "Name").

**Pros:**
- users wouldn't notice the change, it should be 100% seamless
- no need to add extra options
- consistent with the UI, where column labeled "Name" contains titles in taxonomy tables

**Cons:**
- name isn't the same as title and it might not feel natural to update as:
`hammer location update --name 'emea/brno' --new-name 'krno'`
which would then be displayed as 'emea/krno'

Prior to merging this PR I'd like to discuss which of the solutions is better from user's perspective.

ares:
I prefer what's suggested in this PR, --name 'emea/brno' is very confusing and inconsistent with webui. It seems like hostgroups, which has the same problem, works like this PR.

ares:
One thing I find confusing is that non-editable attributes are silently ignored. I'd expect it yells at me like any other unknown attribute
```sh
hammer organization update --id 13 --title some
Organization updated
```

vs

```sh
hammer organization update --id 13 --skljfsldkfj ssdlfj
Could not update the organization:
  Error: Unrecognised option '--skljfsldkfj'
  
  See: 'hammer organization update --help'
```

also `hammer organization --help` lists title without any specific information about it being non-editable.

It's fair to say it's out of scope for this PR but creating an issue to fix this would be nice I think.

ares:
@tstrachota did the discussion happen somewhere? if not, who should be pulled in?

tstrachota:
@ares outcome of discussion on foreman-devel is that we'll use the original approach implemented in this PR. The alternative ways seemed unnatural to other devs.

xprazak2:
The only issue I encountered was with the title update failure in silence (as already mentioned in [previous comment](https://github.com/theforeman/hammer-cli-foreman/pull/299#issuecomment-295696697)), otherwise 👍

tstrachota:
@ares @xprazak2 thanks for the review. 

I filed the "title update failure" here: http://projects.theforeman.org/issues/19598
In this context titles aren't editable at all, but it's a common issue across the whole hammer. It doesn't warn when there's multiple identifier options used together.

PR #300, opened 2017-04-05
Fixes #19172 - Add description to subnet

tbrisker:
depends on https://github.com/theforeman/foreman/pull/4436

### DIFF
--- a/lib/hammer_cli_foreman/subnet.rb
+++ b/lib/hammer_cli_foreman/subnet.rb
@@ -21,6 +21,7 @@ module HammerCLIForeman
     class InfoCommand < HammerCLIForeman::InfoCommand
 
       output ListCommand.output_definition do
+        field :description, _("Description"), Fields::LongText, :hide_blank => true
         field :priority, _("Priority")
         field :dns, _("DNS"), Fields::Reference, :details => :url
         field :dns_primary, _("Primary DNS")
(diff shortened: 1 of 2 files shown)


lib/hammer_cli_foreman/subnet.rb
```diff
@@ -9,6 +9,7 @@ class ListCommand < HammerCLIForeman::ListCommand
       output do
         field :id, _("Id")
         field :name, _("Name")
+        field :description, _("Description")
```
ohadlevy: just a general query, if i use this version of hammer against an older version of foreman, does this work as expected? e.g. the description header is not shown or its shown but its value is always empty?

tstrachota: Unfortunately it will always show empty column. It is possible to use 
```ruby
field :description, _("Description"), Fields::Field, :hide_blank => true
```
but it's not completely the same plus it won't work in lists.

If we wanted to implement hiding unsupported fields we would probably need to rely on `foreman_version` from response headers, which is more effective than checking if the key is missing in the whole list.

tstrachota:
I tested the commands a bit more. Server allows to store multiline values. In such case it breaks table output. I wanted to fix that but unfortunately I found out it's currently blocked by [#19191](http://projects.theforeman.org/issues/19191).

Therefore I suggest either to remove the column from list and use `Fields::LongText, :hide_blank => true` field type or format descriptions in `extend_data(subnet)`.

tbrisker:
@tstrachota I hope I understood your suggestion correctly, please re-review.

tstrachota:
Works fine now:
```
Id:            2
Name:          tstracho-laptop
Network:       192.168.122.0
Mask:          255.255.255.0
VLAN ID:       
Description:   
  this
  is
  description
  test
...
```
Thanks @tbrisker

PR #301, opened 2017-04-07
Fixes #18742 - correctly resolve empty arrays

tstrachota:


lib/hammer_cli_foreman/id_resolver.rb
```diff
@@ -144,7 +144,13 @@ def get_id(resource_name, options)
     end
 
     def get_ids(resource_name, options)
-      options[HammerCLI.option_accessor_name("ids")] || find_resources(resource_name, options).map{|r| r['id']}
+      if options[HammerCLI.option_accessor_name("ids")]
+        options[HammerCLI.option_accessor_name("ids")]
```
ares: this might be slighly less repetetive and a bit more readable (first two lines)

```ruby
if (ids = options[HammerCLI.option_accessor_name("ids")])
  ids
elsif !options_empty?(@api.resource(resource_name), options)
  find_resources(resource_name, options).map{ |r| r['id'] }
else
  []
end
```

if you agree, the same could be applied below on line 167

tstrachota:
@ares good idea, updated

tstrachota:
Merged, thanks for the review @ares

PR #302, opened 2017-04-19
Fixes #19312 - use Array for empty attributes

ares:
the method sends attributes in Array form if some are present


tstrachota:
:+1: fixes the issue for me. Thanks @ares

PR #303, opened 2017-04-19
Fixes #14872 - prevent sending nil values in hostgroup update

tstrachota:


ares:
Looks good, fixes the issue in my environment, ACK

PR #305, opened 2017-04-24
Fixes #19362 - custom error for 401 unauthorized

tstrachota:
Previous solution modified message of the original `RestClient::Unauthorized`. Unfortunatly this attribute isn't writable in all RestClient versions.

Requires: https://github.com/Apipie/apipie-bindings/pull/64


mbacovsky:
:+1: Looks good, thanks @tstrachota

PR #307, opened 2017-04-26
fixes #4396 - add additional attributes on user list command.

ohadlevy:
Align output with UI to additionally include:
- Admin Flag
- Last Login timestamp
- Authorized By


ares:
one nit, otherwise works well

ares:
I don't have permissions here but works for me and code looks correct, ACK from my side

tstrachota:
Merged, thanks @ohadlevy @ares

PR #309, opened 2017-05-02
Fixes #19431 - retry command on session expiry

tstrachota:
Requires: https://github.com/theforeman/hammer-cli/pull/238

### DIFF
--- a/lib/hammer_cli_foreman/api.rb
+++ b/lib/hammer_cli_foreman/api.rb
@@ -3,6 +3,7 @@ require 'hammer_cli_foreman/api/connection'
 module HammerCLIForeman
   module Api
     class UnauthorizedError < RuntimeError; end
+    class SessionExpired < UnauthorizedError; end
   end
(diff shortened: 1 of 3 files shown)


ares:
does this mean that every hammer plugin requires similar patch?

tstrachota:
@ares no, as long as the exception handler in the plugin inherits from the foreman one (hammer-cli-katello does, I didn't check the rest). By default foreman exception handler is used in plugins, so it's safe unless the plugin overrides exception handler with custom one.

ares:
ok cool, I think I "misgrepped" it yesterday, foreman openscap seems to inherit it [properly](https://github.com/theforeman/hammer_cli_foreman_openscap/blob/1d8465f8ce74d6952a8b8ae67564143ae9dff17f/lib/hammer_cli_foreman_openscap/exception_handler.rb#L2)

thanks, looks good to me (and works :+1:)

PR #310, opened 2017-05-05
Fixes #19470 - provide default exception handler

tstrachota:


tstrachota:
There's one related PR into core https://github.com/theforeman/hammer-cli/pull/239 that adds logging of what handler is used.

mbacovsky:
:+1: Looks good. I didn't found any cases where this could break things.

tstrachota:
Thanks, @mbacovsky, merging.

PR #312, opened 2017-06-19
Fixes #17135 - Adding a list normalizer to override-value-order

orrabin:


mbacovsky:
@orrabin the code looks good. The only issue (usability) is that the list normalizer is quite picky about what kind of quoting is used:
`hammer sc-param update --override true  --id 32  --override-value-order 'fqdn,datacentre,region,lifecycle_environment,hostgroup,"datacentre,lifecycle_environment"'` resulted in proper data in the UI:
```
fqdn
datacentre
region
lifecycle_environment
hostgroup
datacentre,lifecycle_environment
```
while `hammer sc-param update --override true  --id 32  --override-value-order "'fqdn','datacentre','region','lifecycle_environment','hostgroup','datacentre,lifecycle_environment'"` proceeded sucessfully but the result was
```
'fqdn'
'datacentre'
'region'
'lifecycle_environment'
'hostgroup'
'datacentre
lifecycle_environment'
```
and one more take `hammer sc-param update --override true  --id 32  --override-value-order fqdn,datacentre,region,lifecycle_environment,hostgroup,"datacentre,lifecycle_environment"` with the result
```
fqdn
datacentre
region
lifecycle_environment
hostgroup
datacentre
lifecycle_environment
```

Any idea how to make it less error prone for the users? I guess it is more about how the list parser works, but I have no idea how to improve it.

orrabin:
@mbacovsky thanks for the review.

I can think of two options here:
1) change documentation to state this expects the format `fqdn\nhostgroup\nos`
2) remove the normalizer and document this should be sent as an array - this will create a difference from the api that does not accept arrays

Changing the parser would of course also work but it will likely still be error prone when mixing between different quotes.
Maybe we should ignore the first quotes and treat: `'hostgroup,"datacentre,lifecycle_environment"'` as if it were sent as: `hostgroup,"datacentre,lifecycle_environment"`. Then any quotes inside would be used to identify when not to split on a comma.
I'm not sure that would solve all three cases but it might be a little friendlier.
The other option is to use something other then quotes to identify what needs to stay together but I think each person will have a different idea on what is the correct character to use.

mbacovsky:
@orrabin, I'm thinking about fixing the list normalizer as this affects every option of type array. Another issue is that unexpected quotation doesn't rise an error but proceeds with unexpected result. I'll do some tests and let you know.

mbacovsky:
@orrabin, could you take a look at https://github.com/theforeman/hammer-cli/pull/246. The new CSV parser should fix the usability issues. I'd be glad for any comments or concerns about the CSV syntax.

orrabin:
@mbacovsky I responded on the other pr, looks good with what I tested.

test/functional/smart_variable_test.rb
```diff
@@ -10,6 +10,7 @@
 
     api_expects(:smart_variables, :update, 'Update the variable') do |par|
       par['smart_variable']['variable'] == 'name2'
+      par['smart_variable']['override_value_order'] = 'fqdn,hostgroup,domain,os'
```
tstrachota: There's `&&` missing, to make sure both lines influence result of the test. The latter line is assignment now -> the test will always succeed.

lib/hammer_cli_foreman/smart_variable.rb
```diff
@@ -84,27 +93,32 @@ class CreateCommand < HammerCLIForeman::CreateCommand
...
-      build_options :without => [:variable_type, :validator_type]
+      build_options :without => [:variable_type, :validator_type, :override_value_order]
 
       option "--variable-type", "VARIABLE_TYPE", _("Type of the variable."),
         :format => HammerCLI::Options::Normalizers::Enum.new(
             ['string', 'boolean', 'integer', 'real', 'array', 'hash', 'yaml', 'json'])
       option "--validator-type", "VALIDATOR_TYPE", _("Type of the validator."),
         :format => HammerCLI::Options::Normalizers::Enum.new(['regexp', 'list', ''])
+      option "--override-value-order", "OVERRIDE_VALUE_ORDER", _("The order in which values are resolved"),
+            :format => HammerCLI::Options::Normalizers::List.new
```
tstrachota: Aren't we missing `include SmartVariableUpdateCreateCommons` for the create command here?

lib/hammer_cli_foreman/smart_variable.rb
```diff
@@ -84,27 +93,32 @@ class CreateCommand < HammerCLIForeman::CreateCommand
...
-      build_options :without => [:variable_type, :validator_type]
+      build_options :without => [:variable_type, :validator_type, :override_value_order]
 
       option "--variable-type", "VARIABLE_TYPE", _("Type of the variable."),
         :format => HammerCLI::Options::Normalizers::Enum.new(
             ['string', 'boolean', 'integer', 'real', 'array', 'hash', 'yaml', 'json'])
       option "--validator-type", "VALIDATOR_TYPE", _("Type of the validator."),
         :format => HammerCLI::Options::Normalizers::Enum.new(['regexp', 'list', ''])
+      option "--override-value-order", "OVERRIDE_VALUE_ORDER", _("The order in which values are resolved"),
+             :format => HammerCLI::Options::Normalizers::List.new
```
tstrachota: Nitpick: the option definition could be moved to `SmartVariableUpdateCreateCommons` too (as well as the other common options above).

orrabin:
@tstrachota thanks, I addressed all the commits. Should I leave 1 test for update or write my own separate test?

tstrachota:
@orrabin ty, I think it's fine just to extend the condition in the `api_expects` block

orrabin:
@tstrachota I added the `override-value-order` in the test.

tstrachota:
Merged, thanks @orrabin

PR #320, opened 2017-07-17
Fixes #20315 - Sessions mutually exclusive with basic auth

tstrachota:
It fixes http://projects.theforeman.org/issues/20131 too.

### DIFF
--- a/lib/hammer_cli_foreman/api/session_authenticator_wrapper.rb
+++ b/lib/hammer_cli_foreman/api/session_authenticator_wrapper.rb
@@ -18,6 +18,7 @@ module HammerCLIForeman
 
       def clear
         destroy_session
+        @authenticator.clear if @authenticator.respond_to?(:clear)
       end
 
       def status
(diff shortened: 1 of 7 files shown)


lib/hammer_cli_foreman/api/session_authenticator_wrapper.rb
```diff
@@ -28,13 +29,22 @@ def status
         end
       end
 
+      def force_user_change
+        @user_changed = true
+      end
+
+      def user_changed?
+        @user_changed
```
ares: usually `!!@user_changed` is used to ensure the result of method with `?` at the end returns boolean

tstrachota:
@ares the issues you're mentioning should actually be the main thing this PR is supposed to fix. I couldn't reproduce them. Can you double-check you checked out correct branch and if yes, provide your config files?

tstrachota:
Rebased

ares:
thanks for helping to debug, the `bundle update` helped, now it works as expected, could you just make a note in example foreman.yml that if user enables sessions, stored credentials are ignored? otherwise :+1:

tstrachota:
@ares I added a comment to the default config file and fixed using incorrect password when sessions are off and a different username is passed.

ares:
:+1: works for me

PR #323, opened 2017-08-14
Fixes #20575 - Don't store session_id for unauthorized responses

tstrachota:


xprazak2:
I am not entirely sure how this could be reproduced, but anyway.
I allowed sessions in hammer config, then:

```
> hammer -u admin -p correct_pwd host list
...
> hammer auth status
Session exists, currently logged in as 'admin'
> hammer auth logout
Logged out.
> hammer -u admin -p wrong_pwd host list
401 Unauthorized
> hammer auth status
Using sessions, you are currently not logged in.
```

When trying to log in with incorrect password, the debug output shows there is a ` _session_id` in response headers:
```
...
:set_cookie => [
        [0] "_session_id=c55bb0c4fd93eb43cd8ed80d90c0fdf1; path=/; HttpOnly"
    ],
...
```

But the session is clearly not established because of a newly added [condition](https://github.com/theforeman/hammer-cli-foreman/pull/323/files#diff-665011da71e3427717916be197b0a6faR74), which means session is not saved on unauthorized request.

xprazak2:
👍

tstrachota:
Thanks for the review @xprazak2

PR #325, opened 2017-08-23
Fixes #20712 - REX option would be show if its enabled on interface

snagoor:
REX option would show for the interface if its enabled. See below 

~~~
hammer host info --id 8
Id:                       8
Name:                     rhel70.lab.example.com
Organization:             Default Organization
Location:                 Default Location
Puppet CA Id:             
Puppet Master Id:         
Cert name:                rhel70.lab.example.com
Managed:                  no
Installed at:             
Last report:              
Network:                  
    IP:  192.168.121.166
    MAC: 52:54:00:62:a3:74
Network interfaces:       
 1) Id:          10
    Identifier:  eth0
    Type:        interface (primary, provision, execution)    -----> execution option is shown 
    MAC address: 52:54:00:62:a3:74
    IP address:  192.168.121.166
    FQDN:        rhel70.lab.example.com
~~~

### DIFF
--- a/lib/hammer_cli_foreman/interface.rb
+++ b/lib/hammer_cli_foreman/interface.rb
@@ -9,6 +9,7 @@ module HammerCLIForeman
       flags = []
       flags << _('primary') if nic['primary']
       flags << _('provision') if nic['provision']
+      flags << _('execution') if nic['execution']
 
       type = nic['type']
       if !flags.empty?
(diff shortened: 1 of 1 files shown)


snagoor:
this depends on PR https://github.com/theforeman/foreman/pull/4781

tstrachota:
Thanks for the patch, @snagoor. Similar to the related API PR I think this should be an extension from hammer-cli-remote-execution plugin. 
You can find an example of such code in katello: https://github.com/Katello/hammer-cli-katello/blob/master/lib/hammer_cli_katello/host_extensions.rb#L19

Unfortunately it's currently possible to only append to the output and not insert in the middle. There's starting to be more and more similar calls for output modification from plugins. Maybe it's the right time to re-evaluate and provide some interface for that in hammer-core.

tstrachota:
I do think this should be implemented in the remote execution CLI plugin. An API that will enable this should be soon available: https://github.com/theforeman/hammer-cli/pull/275

ofedoren:
@snagoor, ping.

snagoor:
@ofedoren, sorry closing this PR and will open another one in the respective project

PR #326, opened 2017-08-25
Fixes #18750 - added fields info which absent in hostgroup info

kgaikwad:
Foreman [PR]<https://github.com/theforeman/foreman/pull/4776> which is needed for this change is merged now.
Not created separate label for "Activation-Keys" as it is already listed under parameters label.

### DIFF
--- a/lib/hammer_cli_foreman/hostgroup.rb
+++ b/lib/hammer_cli_foreman/hostgroup.rb
@@ -61,7 +61,7 @@ module HammerCLIForeman
         field :name, _("Name")
         field :title, _("Title")
         field nil, _("Operating System"), Fields::SingleReference, :key => :operatingsystem
-        field nil, _("Environment"), Fields::SingleReference, :key => :environment
+        field nil, _("Puppet Environment"), Fields::SingleReference, :key => :environment
         field nil, _("Model"), Fields::SingleReference, :key => :model
       end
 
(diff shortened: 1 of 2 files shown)


lib/hammer_cli_foreman/hostgroup.rb
```diff
@@ -61,7 +61,7 @@ class ListCommand < HammerCLIForeman::ListCommand
         field :name, _("Name")
         field :title, _("Title")
         field nil, _("Operating System"), Fields::SingleReference, :key => :operatingsystem
-        field nil, _("Environment"), Fields::SingleReference, :key => :environment
+        field :environment_name, _("Puppet Environment")
```
tstrachota: Please use
```ruby
field nil, _("Puppet Environment"), Fields::SingleReference, :key => :environment
```
for the reason of consistency.

lib/hammer_cli_foreman/hostgroup.rb
```diff
@@ -73,19 +73,28 @@ class InfoCommand < HammerCLIForeman::InfoCommand
...
       output ListCommand.output_definition do
         field :description, _("Description"), Fields::LongText, :hide_blank => true
-        field nil, _("Subnet"), Fields::SingleReference, :key => :subnet
-
-        field nil, _("Domain"), Fields::SingleReference, :key => :domain
-        field nil, _("Architecture"), Fields::SingleReference, :key => :architecture
-        field nil, _("Partition Table"), Fields::SingleReference, :key => :ptable
-        field nil, _("Medium"), Fields::SingleReference, :key => :medium
+        field :ancestry, _("Parent Id"), Fields::Field, :hide_blank => true
+        field :parent_name, _("Parent Name"), Fields::Field, :hide_blank => true
```
tstrachota: You can use
```ruby
field nil, _("Parent"), Fields::SingleReference, :key => :parent
```
here and remove the ancestry field. Parent id is returned also in `:parent_id`.

lib/hammer_cli_foreman/hostgroup.rb
```diff
@@ -73,19 +73,28 @@ class InfoCommand < HammerCLIForeman::InfoCommand
...
         field :puppet_ca_proxy_id, _("Puppet CA Proxy Id")
+        field :puppet_ca_proxy_name, _("Puppet CA Proxy")
```
tstrachota: Ditto `SingleReference`

tstrachota:
Thanks for the PR @kgaikwad 
Please replace all the `*_name` fields with references. I added some examples in the incline comments.

When you're in the code, would you mind extending also host info? The changes will be nearly the same.

kgaikwad:
@tstrachota ,

Yeah, sure. I will add change for `host` also.
Could you please share `issue link` for reference for `host` changes? 
or Should I go and make changes as per host tabs web-ui for sequence and each tab's fields.

tstrachota:
@kgaikwad thanks. I updated the original issue and added some more info about hosts. Basically we need to go through the host/hostgroup form and make sure CLI provides the same information.

BTW I found out the original issue description didn't mention "PXE loader" field that is missing too.

kgaikwad:
@tstrachota ,
Thank you, I will update PR with new changes.

PR #328, opened 2017-08-28
Fixes #20758 - Show empty table field with ReferenceFormatter

xprazak2:
When [data_for_field](https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/output/adapter/table.rb#L78) are `nil`, empty string is passed to [format](https://github.com/theforeman/hammer-cli-foreman/blob/master/lib/hammer_cli_foreman/output/formatters.rb#L37)

### DIFF
--- a/lib/hammer_cli_foreman/output/formatters.rb
+++ b/lib/hammer_cli_foreman/output/formatters.rb
@@ -35,7 +35,7 @@ module HammerCLIForeman::Output
       end
 
       def format(reference, field_params={})
-        return "" if reference.nil?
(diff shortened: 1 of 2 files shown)


tstrachota:
:+1: thanks, @xprazak2 Would you mind extending https://github.com/theforeman/hammer-cli-foreman/blob/master/test/unit/output/formatters_test.rb#L46 with a simple test so that we have this case covered?

xprazak2:
I added a test covering this.

tstrachota:
:+1: thanks @xprazak2 !

PR #329, opened 2017-09-01
Fixes #20476 - Add ssh-keys commands

bastilian:
The pull request will add commands to hammer to manage users' ssh-keys with the following commands:

```
hammer user ssh-keys list ...
hammer user ssh-keys add ...
hammer user ssh-keys delete ...
```
Two things I would need some guidance on:
- The commands are implemented, but still need to be made subcommands of 'user'. 
- Tests are based on what is in other unit tests, but fail and I can't make out what is still missing to make it work.

### DIFF
--- a/lib/hammer_cli_foreman/ssh_keys.rb
+++ b/lib/hammer_cli_foreman/ssh_keys.rb
@@ -1,0 +1,47 @@
+module HammerCLIForeman
+  class SSHKeys < HammerCLIForeman::Command
+    resource :ssh_keys
+    command_name "ssh-keys"
+    desc _("Manage SSH Keys sources.")
+
+    class ListCommand < HammerCLIForeman::ListCommand
+      output do
+        field :id, _("Id")
+        field :name, _("Name")
+        field :fingerprint, _("Fingerprint")
+        field :length, _("Length")
+        field :created_at, _("Created at"), Fields::Date
+      end
+
+      build_options
+    end
+
+    class CreateCommand < HammerCLIForeman::CreateCommand
+      command_name "add"
+
+      option "--key-file", "KEY_FILE", _("Path to a SSH public key"),
+             :format => HammerCLI::Options::Normalizers::File.new
+
+      def request_params
+        params = super
+        params['user_id'] ||= params['ssh_key']['user_id']
+        params['ssh_key']['key'] ||= options['option_key_file']
+        params
+      end
+
+      success_message _("SSH Key %{name} added")
+      failure_message _("Could not add SSH Key")
+
+      build_options
+    end
+
+    class DeleteCommand < HammerCLIForeman::DeleteCommand
+      success_message _("SSH Key deleted")
+      failure_message _("Could not delete the SSH Key")
+
+      build_options
+    end
+
+    autoload_subcommands
+  end
+end
--- a/lib/hammer_cli_foreman/user.rb
+++ b/lib/hammer_cli_foreman/user.rb
@@ -135,10 +135,10 @@ module HammerCLIForeman
     end
 
     HammerCLIForeman::AssociatingCommands::Role.extend_command(self)
-
+    lazy_subcommand('ssh-keys', _("Managing User SSH Keys."),
+      'HammerCLIForeman::SSHKeys', 'hammer_cli_foreman/ssh_keys'
+    )
     autoload_subcommands
   end
 
 end
-
-
--- a/test/functional/ssh_keys_test.rb
+++ b/test/functional/ssh_keys_test.rb
@@ -1,0 +1,89 @@
+require File.join(File.dirname(__FILE__), 'test_helper')
+
+describe 'ssh_keys' do
+  let(:base_cmd) { ['user', 'ssh-keys'] }
+  let(:user) do
+    {
+     :id => 1,
+     :name => 'admin'
+    }
+  end
+  let(:ssh_key) do
+    {
+      :id => 1,
+      :user_id => user[:id],
+      :name => 'SSH Key',
+      :fingerprint => 'FINGERPRINT',
+      :key => 'KEY',
+      :lenght => 256
+    }
+  end
+
+  describe 'list' do
+    let(:cmd) { base_cmd << 'list' }
+
+    it 'lists all ssh-keys for a given user' do
+      params = ['--user-id', user[:id]]
+      api_expects(:ssh_keys, :index, :user_id => user[:id])
+        .returns(index_response([ssh_key]))
+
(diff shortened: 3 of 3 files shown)


lib/hammer_cli_foreman/ssh_keys.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    class DeleteCommand < HammerCLIForeman::DeleteCommand
+      include SSHKeyCommons
+
+      success_message _("SSH Key [%{name}] deleted")
+      failure_message _("Could not delete the SSH Key")
+
+      build_options
+    end
+end
```
adamruzicka: Missing an end here `hammer-cli-foreman/lib/hammer_cli_foreman/ssh_keys.rb:54: syntax error, unexpected end-of-input, expecting keyword_end (SyntaxError)`

bastilian: bad commit. fixed. :)

lib/hammer_cli_foreman.rb
```diff
@@ -107,6 +107,10 @@ def self.exception_handler_class
       'HammerCLIForeman::SmartClassParameter', 'hammer_cli_foreman/smart_class_parameter'
     )
 
+    HammerCLI::MainCommand.lazy_subcommand('ssh-keys', _("Manipulate smart class parameters."),
```
ares: I hope it's not manipulating smart class parameters ;-)

bastilian: It's "Managing User SSH Keys.". corrected. :)

bastilian:
Thanks to @adamruzicka the ssh-keys commands are now subcommands of 'user'

lib/hammer_cli_foreman/ssh_keys.rb
```diff
@@ -0,0 +1,57 @@
...
+        field :name, _("Name")
+        field :fingerprint, _("Fingerprint")
+        field :length, _("Length")
+        field :created_at, _("Created at"), Fields::Date
+      end
+
+      build_options
+    end
+
+    class CreateCommand < HammerCLIForeman::CreateCommand
```
tstrachota: It would be good to add an option `--key-file` that would take a path to the pub key, read it from there and pass it to the `key` parameter. It should be easy to add with https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/options/normalizers.rb#L134

bastilian:
@tstrachota moved the tests now to functional tests. They still require the 1.15 api test data to be updated to incliude, as it should. Which should probably be done in a separate PR, or?

 I'll implement the `--key-file` argument later today.

lib/hammer_cli_foreman/ssh_keys.rb
```diff
@@ -0,0 +1,57 @@
+module HammerCLIForeman
+  module SSHKeyCommons
+    def self.included(base)
+      base.option "--user_id", "USER_ID", _("ID of the user to add the SSH Key to")
```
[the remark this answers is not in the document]
bastilian: I removed SSHKeyCommons now.

lib/hammer_cli_foreman/ssh_keys.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+    def request_params
+      params = super
+      params["user_id"] = option_user_id
```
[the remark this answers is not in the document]
bastilian: SSHKeyCommons was overriding the default behavior. `--user-id` and `--user` work now.

lib/hammer_cli_foreman/ssh_keys.rb
```diff
@@ -0,0 +1,57 @@
...
+        field :created_at, _("Created at"), Fields::Date
+      end
+
+      build_options
+    end
+
+    class CreateCommand < HammerCLIForeman::CreateCommand
+      include SSHKeyCommons
+
+      def self.command_name
```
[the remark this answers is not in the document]
bastilian: :white_check_mark: check.

lib/hammer_cli_foreman/ssh_keys.rb
bastilian: Check.

bastilian:
@tstrachota thank you for your review and help! The add command now also accepts a `--key-file` argument to pass a key.

lib/hammer_cli_foreman/ssh_keys.rb
```diff
@@ -0,0 +1,41 @@
...
+      end
+
+      build_options
+    end
+
+    class CreateCommand < HammerCLIForeman::CreateCommand
+      command_name "add"
+
+      option "--key-file", "KEY", _("Path to a SSH public key"),
+             :attribute_name => :option_key,
```
tstrachota: I found one tricky thing. When you map two options to the same attribute name the latter attribute uses the reader that was already defined. This prevents from using two options with different formatters pointing to the same `:attribute_name`. You'll have to keep the attribute names separate and handle it in `request_params`.

Something like this should fix it:
```ruby
def request_params
  params = super
  params['ssh_key']['key'] ||= options['option_key_file']
  params
end
```

lib/hammer_cli_foreman/ssh_keys.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    class CreateCommand < HammerCLIForeman::CreateCommand
+      command_name "add"
+
+      option "--key-file", "KEY_FILE", _("Path to a SSH public key"),
+             :format => HammerCLI::Options::Normalizers::File.new
+
+      def request_params
+        params = super
+        params['user_id'] ||= params['ssh_key']['user_id']
```
bastilian: This was needed to make it work with a real API, in the tests the the expectation of the `user_id` param gets satisfied, but when using the CLI the id gets wrapped into the ssh_key hash. I'm not sure if that is an issue to do with the API or the parameter resolving.

bastilian:
@tstrachota I've made the changes you suggested. Also, see line comment. :)

tstrachota:
I added a comment about the apidoc issue and merged as: https://github.com/theforeman/hammer-cli-foreman/commit/d04e32f880d05573630baef77e56d380fac3b2a7

Thanks @bastilian !

ares:
Did I miss something or there's no way to show the actual key? Should it be added?

bastilian:
@ares You mean to show the actual public key? A show comman would only show what the list command already does, but if the public key should also be shown then it makes more sense to have a separate command for a single key.

ares:
Yeah I mean public key. It's nice to see fingerprint, but I'd expect to see all information via show command which seems to be missing.

PR #331, opened 2017-09-06
Fixes #20754 - remove --include option from host list

kgaikwad:


tstrachota:
Thank you @kgaikwad. Can you please do the same for `--thin` parameter on host list and `--include` on hostgroup list? It's the same issue so I think we can fix it in this PR.

kgaikwad:
@tstrachota ,
Updated the PR as well as an issue with suggested changes.

tstrachota:
Thank you @kgaikwad !

PR #332, opened 2017-09-14
Update apidoc for 1.15, add 1.16

tstrachota:
Updates apipie cache in tests for latest foreman 1.15 and adds export for 1.16. I'm keeping 1.15 as default until the next version is released.

I removed one test that is no longer relevant. It was fulfilling the failure expectation for a different error anyway.


mbacovsky:
:+1: thanks @tstrachota!

PR #333, opened 2017-09-25
Refs #21013 - Test to expect integer params as a number

mbacovsky:
This patch is fixing the tests to expect correct parameter type for Integers


tstrachota:
:+1: thanks @mbacovsky

PR #334, opened 2017-10-09
Refs #20476 - Add show command for ssh-keys

bastilian:
A follow up on #329 to add a show command for ssh-keys that includes the public key as well.

Example:
```shell
$ hammer user ssh-keys show --user-id 4 --id 1
Id:          1
Name:        bastilian@bastilian.local
Fingerprint: c1:e0:a7:7c:06:7d:f5:c3:po:r2:d2:d7:a5:0e:9b:42
Length:      256
Created at:  2017/09/19 08:53:35
Public Key:  ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAINDirKN2unK1h/3sXNp7KsCbwF86YLVPJfGLgwDjrnZI bastilian@bastilian.local
```

### DIFF
--- a/lib/hammer_cli_foreman/ssh_keys.rb
+++ b/lib/hammer_cli_foreman/ssh_keys.rb
@@ -36,6 +36,14 @@ module HammerCLIForeman
       build_options
     end
 
+    class InfoCommand < HammerCLIForeman::InfoCommand
+      output ListCommand.output_definition do
+        field :key, _("Public Key")
+      end
+
+      build_options
+    end
+
     class DeleteCommand < HammerCLIForeman::DeleteCommand
       success_message _("SSH Key deleted")
       failure_message _("Could not delete the SSH Key")
--- a/test/functional/ssh_keys_test.rb
+++ b/test/functional/ssh_keys_test.rb
@@ -71,6 +71,22 @@ describe 'ssh_keys' do
     end
   end
 
+  describe 'info' do
+    let(:cmd) { base_cmd << 'info' }
+
+    it 'shows the public key' do
+      params = ['--id', ssh_key[:id],
+                '--user-id', user[:id]]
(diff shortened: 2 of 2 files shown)


bastilian:
@ares mind taking if this does what you'd expect from the show command?

ares:
It works just fine, one thing, while I find `show` better, we use `info` in hammer for show actions so the command should be renamed. I appologize since I know I asked for show command but it turned out while I was testing it...

lib/hammer_cli_foreman/ssh_keys.rb
```diff
@@ -36,6 +36,16 @@ def request_params
       build_options
     end
 
+    class InfoCommand < HammerCLIForeman::InfoCommand
+      command_name 'show'
```
ares: removing this should be enough

bastilian:
@ares changed. :) show is now info.

ares:
Tested, works great, code also looks good. I hope @mbacovsky or @tstrachota could take a look and merge. :+1: from me.

mbacovsky:
@bastilian So I've found out that the issue is in the API docs where the parameter [1] should be required.
Filed as [2]
[1] https://github.com/theforeman/foreman/blob/develop/app/controllers/api/v2/ssh_keys_controller.rb#L12
[2] http://projects.theforeman.org/issues/21299

mbacovsky:
@bastilian I have fix for Foreman API ready, so I'll test that it works with your PR. If yes, I'll ACK this PR and merge as there is no need to wait for the changes in the API while `user ssh-key delete` is already broken the same way.

mbacovsky:
https://github.com/theforeman/foreman/pull/4907 that solves the `user_id` resolving sent.

:+1: Thanks for the contribution @bastilian and @ares for the review.

PR #335, opened 2017-10-10
Fixes #21264 - Add org/loc in ptable info command

rahulbajaj0509:


tstrachota:
:+1: works fine, thanks @rahulbajaj0509

PR #339, opened 2017-11-08
Fixes #20742 - Make hammer host reports work

xprazak2:
I tried to make this fix as simple as possible and contain it in the report subcommand. Needs additional [search](https://github.com/theforeman/foreman/pull/4989).

### DIFF
--- a/lib/hammer_cli_foreman/commands.rb
+++ b/lib/hammer_cli_foreman/commands.rb
@@ -212,8 +212,12 @@ module HammerCLIForeman
 
       # resolve 'id' parameter if it's defined as an option
       id_option_name = HammerCLI.option_accessor_name('id')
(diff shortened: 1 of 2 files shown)


ares:
Works well, fixed the problem :+1:

mbacovsky:
@xprazak2, @ares, the change in the abstract command didn't felt right. While investing it I found out some issues in how the report command is done today. I changed the predecessor of the command to ListCommand as that better matches the API routes. In the end I came up with working solution that deviated significantly from this PR. After discussion with @xprazak2 I created new PR (https://github.com/theforeman/hammer-cli-foreman/pull/340) from it. Would it be acceptable replacement of this PR?

xprazak2:
closing in favour of #340

PR #340, opened 2017-11-13
Fixes #20742 - Make hammer host reports work

mbacovsky:
Added proper mapping of id and name options to search queries. Made host
parameter required. I found out there is no name in reports and fixed
the id resolver. --search was fixed. Added some tests.

### DIFF
--- a/lib/hammer_cli_foreman/host.rb
+++ b/lib/hammer_cli_foreman/host.rb
@@ -213,12 +213,33 @@ module HammerCLIForeman
     end
 
 
-    class ReportsCommand < HammerCLIForeman::AssociatedResourceListCommand
+    class ReportsCommand < HammerCLIForeman::ListCommand
       command_name "reports"
       resource :reports, :index
-      parent_resource :hosts
+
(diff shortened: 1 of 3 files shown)


xprazak2:
I tested, works nicely. I will close my PR in favour of this.

lib/hammer_cli_foreman/host.rb
```diff
@@ -213,12 +213,33 @@ def send_request
...
+        validator.any(:option_name, :option_id).required
+      end
+
+      def request_params
+        params = super
+        search = []
+        search << params['search'] if params['search']
+        
+        hostname = get_option_value('name')
+        search << "host=\"#{hostname}\"" if hostname
```
ares: you could use `%Q(host="#{hostname}")` to avoid escaping if you want

ares:
LGTM, left one minor comment

mbacovsky:
@ares, @xprazak2, thanks, updated.

ares:
:+1:

PR #342, opened 2017-11-15
Fixes #21674 - Add abstraction for subcommand searching

xprazak2:


xprazak2:
Limitation:
```
hammer resource-a resource-bs --resource-a-id 13 --some-attribute some_value
```
if I create a subcommand in hammer_cli_plugin_a, where `resourceA` comes from `pluginA` and `resourceB` from core and `pluginC` extends `resourceB` with `some_attribute` that is not a foreign key in `resourceB`, then the search will not work, but `--some-attribute` will still be shown as acceptable option.

ares:
What's the status here? It blocks linked PR

ares:
Could someone please take a look? This blocks further hammer openscap improvements and it's been sitting untouched here more than 2 months.

ares:
any progress?

ares:
ping @theforeman/hammer-admin

lib/hammer_cli_foreman/commands.rb
```diff
@@ -311,6 +311,86 @@ def should_retrieve_all?
...
   end
 
+  class ListSearchCommand < ListCommand
+    def self.search_resource(res, action = :index)
+      resource res, action
+      default_search_options
+    end
+
+    def self.default_search_options
+      option("--#{module_resource.singular_name}-id", "ID", _("%s Id") % module_resource.singular_name)
```
ofedoren: Some module resources could look like ```:ansible_roles```, so this will lead to ```--ansible_role-id``` which is incorrect.

How about using .tr('_', '-') like ```"--#{module_resource.singular_name}-id".tr('_', '-')```? So there is no need to redefine this method in plugins.

lib/hammer_cli_foreman/commands.rb
```diff
@@ -311,6 +311,86 @@ def should_retrieve_all?
...
+      option("--#{module_resource.singular_name}-name", "NAME", _("%s name") % module_resource.singular_name)
```
ofedoren: Same here.

xprazak2:
I made updates as suggested.

ofedoren:
@xprazak2, needs rebase

xprazak2:
Rebased.

ares:
Can we keep this moving? we should get this into 1.18 repos soon since the ansible hammer plugin is already there.

ofedoren:
@ares, just to be clear, I'm not using this approach in the latest version of ansible hammer plugin. By deafult it's waiting for https://github.com/theforeman/foreman_ansible/pull/161 to be merged.

If this is merged before [161](https://github.com/theforeman/foreman_ansible/pull/161), then a new version of ansible hammer plugin needs to be released.

ares:
@ofedoren oh, based on this last comment Don't merge this for now, please. I guess there is another solution via theforeman/hammer-cli-foreman#342.
I thought we don't need it anymore. Right now it seems we can't assigne roles to host/hostgroup. So lets pick whatever solution and make it working early next week. If you decide the PR in foreman_ansible is the way, let me know. I think it's now waiting on contributor but once updated, I can review, merge and do new release.

ofedoren:
@ares are you sure we can't assign roles? I just checked via `hammer host(group) create/update --ansible-role-ids 1,2,3` and it seems to be working.

Since there is no easier way to get list of host's roles than search like `hammer ansible roles list --search host=hostname` I wanted to add simple commands like `hammer host ansible-roles`. This PR or [161](https://github.com/theforeman/foreman_ansible/pull/161) adds that functionality only and both have nothing in common with roles assigning.

ares:
You're right, I was playing with hammer host ansible-roles and always got a reply that server does not support this. But one can use the command you mentioned, thanks. Still it would be nice to get this PR merged :-)

lib/hammer_cli_foreman/commands.rb
```diff
@@ -279,6 +279,86 @@ def should_retrieve_all?
 
   end
 
+  class ListSearchCommand < ListCommand
```
mbacovsky: Could we have some more descriptive name? It took me while until I figured out what this class does. Would make `AssociatedResourceListCommand` more sense?

lib/hammer_cli_foreman/commands.rb
```diff
@@ -279,6 +279,86 @@ def should_retrieve_all?
...
+    def self.search_resource(res, action = :index)
+      resource res, action
+      default_search_options
+    end
+
+    def self.default_search_options
+      option("--#{module_resource.singular_name}-id".tr('_', '-'), "ID", _("%s Id") % module_resource.singular_name)
```
mbacovsky: I tried to convert `host reports` into descendant of this class. The resulting command looks good and easy to read. However in hammer we traditionally use `--id` and `--name` to refer to the "module resource" in this type of commands i.e. `host reports --id` and `host reports --name`. Would you mind to change the option names accordingly? It may make sense to add a mapping to change it if you have a use case for that.

xprazak2:
Updated, I changed the class name and the options `--id`, `--name` are used.

ofedoren:
@xprazak2 could you please check the tests?

xprazak2:
I fixed tests, all green now.

ares:
any update here? @ofedoren / @mbacovsky

lib/hammer_cli_foreman/commands.rb
```diff
@@ -279,6 +279,86 @@ def should_retrieve_all?
...
+      default_search_options
+    end
+
+    def self.default_search_options
+      option("--id", "ID", _("%s Id") % module_resource.singular_name)
+      option("--name", "NAME", _("%s name") % module_resource.singular_name)
+    end
+
+    def self.search_options_mapping(mapping = {})
+      { "name" => module_resource.singular_name,
```
ofedoren: Rubocop style nitpick :)

Doesn't look better?
```
{
  'name' => module_resource.singular_name,
  'id' => "#{module_resource.singular_name}_id"
}.merge(mapping)

**Comment by ofedoren** [comment]
[test]

**Comment by xprazak2** [comment]
Rebased.

@ofedoren, I was looking at `AssociatedResourceListCommand`. The purpose is indeed similar, however from looking at [FactsCommand](https://github.com/theforeman/hammer-cli-foreman/blob/master/lib/hammer_cli_foreman/host.rb#L183), it seems like it expects nested routes:

```
2019-04-12T09:05:12 [I|app|6fc3dd11] Started GET "/api/hosts/38/facts?page=1&per_page=1000&id=38" for ::1 at 2019-04-12 09:05:12 +0000
2019-04-12T09:05:12 [I|app|6fc3dd11] Processing by Api::V2::FactValuesController#index as JSON
2019-04-12T09:05:12 [I|app|6fc3dd11]   Parameters: {"page"=>"1", "per_page"=>"1000", "id"=>"38", "apiv"=>"v2", "host_id"=>"38", "fact_value"=>{}}
```

Policy and host are not nested so the `policy_id` is never used and I get all the hosts, but I only want the hosts associated to policy:

```
2019-04-12T09:05:22 [I|app|f3281aa6] Started GET "/api/hosts?page=1&per_page=1000&id=1&policy_id=1" for ::1 at 2019-04-12 09:05:22 +0000
2019-04-12T09:05:22 [I|app|f3281aa6] Processing by Api::V2::HostsController#index as JSON
2019-04-12T09:05:22 [I|app|f3281aa6]   Parameters: {"page"=>"1", "per_page"=>"1000", "id"=>"1", "policy_id"=>"1", "apiv"=>"v2", "host"=>{}}
```

I need to specify a search somehow, so that only the associated hosts are returned:

```
2019-04-12T09:17:01 [I|app|a4861261] Started GET "/api/hosts?page=1&per_page=1000&id=1&search=compliance_policy_id%3D1" for ::1 at 2019-04-12 09:17:01 +0000
2019-04-12T09:17:01 [I|app|a4861261] Processing by Api::V2::HostsController#index as JSON
2019-04-12T09:17:01 [I|app|a4861261]   Parameters: {"page"=>"1", "per_page"=>"1000", "id"=>"1", "search"=>"compliance_policy_id=1", "apiv"=>"v2", "host"=>{}}
```

@mbacovsky, I had to revert the option naming back, having just `--id` works fine, but `--name` causes problems. The main reason is that the name is of a parent resource, but hammer uses it to [fetch](https://github.com/theforeman/hammer-cli-foreman/blob/master/lib/hammer_cli_foreman/option_sources/self_param.rb#L11) id of a child.

File: test/unit/commands_test.rb [hunk 6]
```diff
@@ -133,6 +133,30 @@ class Assoc < HammerCLIForeman::AddAssociatedCommand
...
+          command_name 'hosts'
+          search_resource :hosts
+
+          output HammerCLIForeman::Host::ListCommand.output_definition
+
+          build_options
+        end
+      end
+      comm = DomainOuter::HostsCommand.new("", { :adapter => :csv, :interactive => false })
+      out, err = capture_io { comm.run(["--host-id=5"]) }
```
ofedoren: `--host-id` is wrong parameter in this case (it just doesn't exist, since parent command has `domains` resource, so it should be rather `--domain-id` as you designed).

test/unit/commands_test.rb
```diff
@@ -133,6 +133,30 @@ class Assoc < HammerCLIForeman::AddAssociatedCommand
...
+      out.must_equal "Id,Name,Operating System,Host Group,IP,MAC\n2,random-host,\"\",\"\",192.168.100.112,6e:4b:3c:2c:8a:0a\n"
```
ofedoren: There are no quotes in real output, so test will fail.

xprazak2:
I pushed changes, thanks for suggestions @ofedoren!

ofedoren:
Thanks for being patient, @xprazak2! Finally we are merging it :)

PR #344, opened 2017-12-14
Fixes #14725 - Verbose mode does not work and is inconsistent

ofedoren:
depends on https://github.com/theforeman/hammer-cli/pull/263

### DIFF
--- a/lib/hammer_cli_foreman/output/fields.rb
+++ b/lib/hammer_cli_foreman/output/fields.rb
@@ -5,6 +5,7 @@ module Fields
   class SingleReference < Field
 
     def display?(value)
+      return false unless appropriate_verbosity?
       id_key = "#{parameters[:key]}_id"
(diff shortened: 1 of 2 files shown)


tstrachota:
Would it be possible to provide verbosity in the default context instead of passing it everywhere in the tests explicitly?

ofedoren:
@tstrachota, I've made some changes in hammer core, so there is no more need to pass default verbosity everywhere explicitly. Only a few little changes are needed now.

tstrachota:
@ofedoren if I understand it correctly this PR is no longer needed with the latest changes in https://github.com/theforeman/hammer-cli/pull/263 right?

ofedoren:
@tstrachota correct. This PR is no longer needed.

tstrachota:
Ok, thanks for confirmation, I'm closing it.

PR #345, opened 2017-12-15
Fixes #20607 - Allow hiding fields which are missing from api output

ofedoren:
Depends on https://github.com/theforeman/hammer-cli/pull/256


tstrachota:
@ofedoren could you please elaborate why we need to pass output adapters explicitly in the tests that you changed? I don't understand why the change is needed.

ofedoren:
@tstrachota updated. You was right, there is no need to pass output adapters explicitly, that was a my mistake in implementation core changes.

tstrachota:
[test]

tstrachota:
Based on recent changes in the hammer-cli counterpart PR, this is no longer needed. Closing.
@ofedoren thanks for your efforts to keep the plugins compatible.

PR #347, opened 2017-12-28
Fixes #21958 - Add missing id options for loc/org in discovery rules

rahulbajaj0509:
To test, you can run:
$ hammer -u admin -p changeme discovery-rule create --hostgroup-id 1 --location-ids=2 --name "rplevka_dr" --search "last_report = Today" --organization-ids=1


rahulbajaj0509:
Closing this pr and re-opening for theforeman-bot to re-run.

rahulbajaj0509:
@mbacovsky @tstrachota @lzap  the root cause of https://bugzilla.redhat.com/show_bug.cgi?id=1523221 is https://github.com/theforeman/foreman_discovery/pull/386/files here, we have made the `location_ids` and `organization_ids` as `:required => true`, therefore in commands.rb (https://github.com/theforeman/hammer-cli-foreman/blob/master/lib/hammer_cli_foreman/commands.rb#L144) it enters the `if` block which it should not. We need to fix this fast, what is your call?

tstrachota:
@rahulbajaj0509 if the taxonomy ids are required, this is correct behavior. Did you find you why names doesn't get resolved?
The current patch doesn't seem like a correct fix to me.

rahulbajaj0509:
@tstrachota checking with the earlier version of hammer-cli-foreman, to my surprise, the ids were never getting resolved earlier also. 

In the earlier versions, this[https://github.com/rahulbajaj0509/hammer-cli-foreman/blob/master/lib/hammer_cli_foreman/commands.rb#L143-L149] conditional statement was not getting executed as `organization_ids` and `location_ids` were false. Since this condition was not encountered, the values where passed to the API and API handled the rest. But now since these fields are true, we are getting this error. 

Versions in which checking was done:
sat-6211-rhel7 :: ~/Lab/foreman_maintain ‹21293-refactor-disk-io› » rpm -q satellite
satellite-6.3.0-19.0.beta.el7sat.noarch
sat-6211-rhel7 :: ~/Lab/foreman_maintain ‹21293-refactor-disk-io› » rpm -q          
sat-6211-rhel7 :: ~/Lab/foreman_maintain ‹21293-refactor-disk-io› » rpm -qa "*hammer*"                                                                                  
tfm-rubygem-hammer_cli_katello-0.11.0-1.el7sat.noarch
tfm-rubygem-hammer_cli_foreman_tasks-0.0.10.3-1.el7sat.noarch
tfm-rubygem-hammer_cli_foreman_virt_who_configure-0.0.3-1.el7sat.noarch
tfm-rubygem-hammer_cli_foreman_docker-0.0.6-1.el7sat.noarch
tfm-rubygem-hammer_cli-0.11.0.1-1.el7sat.noarch
tfm-rubygem-hammer_cli_foreman_openscap-0.1.5-1.fm1_15.el7sat.noarch
tfm-rubygem-hammer_cli_foreman_bootdisk-0.1.3.3-1.el7sat.noarch
tfm-rubygem-hammer_cli_foreman_remote_execution-0.0.5.3-1.el7sat.noarch
tfm-rubygem-hammer_cli_csv-2.3.0-1.el7sat.noarch
tfm-rubygem-hammer_cli_foreman_discovery-1.0.0-1.el7sat.noarch
tfm-rubygem-hammer_cli_foreman_admin-0.0.7-1.el7sat.noarch
tfm-rubygem-hammer_cli_foreman-0.11.0-1.el7sat.noarch

lzap:
So I suggest to revert the patch which broke it in Satellite 6.3. Then in upstream, let's solve this properly, I suggest to create required option in `def_param_group :taxonomies` if that's allowed by APIPIE. If not, I vote creating new group called `def_param_group_required` with required flags set to true and use that in discovery.

tstrachota:
In short term I agree with reverting the server side patch that introduced the regression as @lzap suggests. That will help us to get rid of the problem until we deliver the proper fix.

The real root cause is in id resolver that ignores `"*_ids"` options:
https://github.com/theforeman/hammer-cli-foreman/blob/master/lib/hammer_cli_foreman/id_resolver.rb#L93

@mbacovsky fixes that in a different patch:
https://github.com/theforeman/hammer-cli-foreman/pull/346/files#diff-5557b50f21382f8aa55f642fe7bda158R98
Once this is merged, hammer should start working even when the taxonomy parameters are marked as required.

rahulbajaj0509:
@tstrachota @lzap thank you so much for your input. Closing this one till then.

PR #353, opened 2018-02-07
Refs #22517 - Do not resolve already resolved id params

mbacovsky:
Avoid resolving of id_parameters that are alredy set.
This allows overriding name resolving by adding
new option_source before the default ones.

### DIFF
--- a/test/unit/option_sources/id_params_test.rb
+++ b/test/unit/option_sources/id_params_test.rb
@@ -1,0 +1,27 @@
+require File.join(File.dirname(__FILE__), '../test_helper')
+
+describe HammerCLIForeman::OptionSources::IdParams do
+  describe "get_options" do
+    class IdParamsTestCommand < HammerCLIForeman::CreateCommand
+      resource :hostgroups
+      build_options
+    end
+
+    let(:hg_cmd) { IdParamsTestCommand.new("", { :adapter => :csv, :interactive => false }) }
+    let(:id_params_source) { HammerCLIForeman::OptionSources::IdParams.new(hg_cmd) }
+
+    it "skips param when set" do
+      hg_cmd.stubs(:get_resource_id).returns(1)
+      option_data = { 'option_domain_id' => 3, 'option_domain_name' => 'test' }
+      params = id_params_source.get_options([], option_data)
+      params.must_equal option_data
+    end
+
+    it "resolves param when unset" do
+      hg_cmd.stubs(:get_resource_id).returns(1)
+      option_data = { 'option_domain_id' => nil, 'option_domain_name' => 'test' }
+      params = id_params_source.get_options([], option_data)
+      params['option_domain_id'].must_equal 1
+    end
+  end
+end
--- a/test/unit/option_sources/id_params_test.rb
+++ b/test/unit/option_sources/id_params_test.rb
@@ -12,7 +12,7 @@ describe HammerCLIForeman::OptionSources::IdParams do
 
     it "skips param when set" do
       hg_cmd.expects(:get_resource_id).with { |res| res.name != :domains }.at_least(0).returns(nil)
-      hg_cmd.expects(:get_resource_id).with { |res| res.name == :domains }.never.returns(1)
+      hg_cmd.expects(:get_resource_id).with { |res| res.name == :domains }.never
       option_data = { 'option_domain_id' => 3, 'option_domain_name' => 'test' }
       params = id_params_source.get_options([], option_data)
       params.must_equal option_data
--- a/test/unit/option_sources/id_params_test.rb
+++ b/test/unit/option_sources/id_params_test.rb
@@ -11,7 +11,7 @@ describe HammerCLIForeman::OptionSources::IdParams do
     let(:id_params_source) { HammerCLIForeman::OptionSources::IdParams.new(hg_cmd) }
 
     it "skips param when set" do
-      hg_cmd.expects(:get_resource_id).with { |res| res.name != :domains }.at_least(0).returns(nil)
+      hg_cmd.stubs(:get_resource_id).returns(nil)
       hg_cmd.expects(:get_resource_id).with { |res| res.name == :domains }.never
       option_data = { 'option_domain_id' => 3, 'option_domain_name' => 'test' }
       params = id_params_source.get_options([], option_data)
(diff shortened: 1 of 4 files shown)


mbacovsky:
@tstrachota, I've added the tests we talked about. All tests are green localy with latest core and this branch rebased.

mbacovsky:
[test]

mbacovsky:
@tstrachota, could you re-check please?

test/unit/option_sources/id_params_test.rb
```diff
@@ -0,0 +1,27 @@
...
+      option_data = { 'option_domain_id' => 3, 'option_domain_name' => 'test' }
+      params = id_params_source.get_options([], option_data)
+      params.must_equal option_data
+    end
+
+    it "resolves param when unset" do
+      hg_cmd.stubs(:get_resource_id).returns(1)
+      option_data = { 'option_domain_id' => nil, 'option_domain_name' => 'test' }
+      params = id_params_source.get_options([], option_data)
+      params['option_domain_id'].must_equal 1
```
tstrachota: Could you test the whole params hash to be sure no other unexpected modifications were made?

test/unit/option_sources/ids_params_test.rb
```diff
@@ -0,0 +1,27 @@
...
+      option_data = { 'option_location_ids' => [3], 'option_location_names' => 'test' }
+      params = ids_params_source.get_options([], option_data)
+      params.must_equal option_data
+    end
+
+    it "resolves param when unset" do
+      cmd.stubs(:get_resource_ids).returns([1])
+      option_data = { 'option_location_id' => nil, 'option_location_names' => 'test' }
+      params = ids_params_source.get_options([], option_data)
+      params['option_location_ids'].must_equal [1]
```
tstrachota: ditto

test/unit/option_sources/id_params_test.rb
```diff
@@ -0,0 +1,27 @@
...
+    class IdParamsTestCommand < HammerCLIForeman::CreateCommand
+      resource :hostgroups
+      build_options
+    end
+
+    let(:hg_cmd) { IdParamsTestCommand.new("", { :adapter => :csv, :interactive => false }) }
+    let(:id_params_source) { HammerCLIForeman::OptionSources::IdParams.new(hg_cmd) }
+
+    it "skips param when set" do
+      hg_cmd.stubs(:get_resource_id).returns(1)
```
tstrachota: This shouldn't be executed in this case, if I understand it correctly.

test/unit/option_sources/id_params_test.rb
```diff
@@ -0,0 +1,27 @@
...
+      option_data = { 'option_domain_id' => 3, 'option_domain_name' => 'test' }
+      params = id_params_source.get_options([], option_data)
+      params.must_equal option_data
+    end
+
+    it "resolves param when unset" do
+      hg_cmd.stubs(:get_resource_id).returns(1)
```
tstrachota: Maybe expects().with()?

mbacovsky:
@tstrachota I've updated the tests. The resource in the testing command has multiple id params. When I set the expectations I had to cover also the params that i'm not working with which may be confusing a bit. Is the new version readable enough?

test/unit/option_sources/ids_params_test.rb
```diff
@@ -0,0 +1,30 @@
...
+      resource :users
+      build_options
+    end
+
+    let(:cmd) { IdsParamsTestCommand.new("", { :adapter => :csv, :interactive => false }) }
+    let(:ids_params_source) { HammerCLIForeman::OptionSources::IdsParams.new(cmd) }
+
+    it "skips param when set" do
+      cmd.expects(:get_resource_ids).with { |res| res.name != :locations }.returns(nil)
+      cmd.expects(:get_resource_ids).with { |res| res.name == :locations }.never.returns([1])
```
tstrachota: Nitpick: isn't `.returns([1])` needless when you expect it to be never called?

test/unit/option_sources/ids_params_test.rb
```diff
@@ -0,0 +1,30 @@
...
+    class IdsParamsTestCommand < HammerCLIForeman::CreateCommand
+      resource :users
+      build_options
+    end
+
+    let(:cmd) { IdsParamsTestCommand.new("", { :adapter => :csv, :interactive => false }) }
+    let(:ids_params_source) { HammerCLIForeman::OptionSources::IdsParams.new(cmd) }
+
+    it "skips param when set" do
+      cmd.expects(:get_resource_ids).with { |res| res.name != :locations }.returns(nil)
```
tstrachota: Nitpick: how come it works without `.at_least(0)` here? I'd expect using it in both tests or nowhere.

test/unit/option_sources/id_params_test.rb
```diff
@@ -0,0 +1,30 @@
...
+    class IdParamsTestCommand < HammerCLIForeman::CreateCommand
+      resource :hostgroups
+      build_options
+    end
+
+    let(:hg_cmd) { IdParamsTestCommand.new("", { :adapter => :csv, :interactive => false }) }
+    let(:id_params_source) { HammerCLIForeman::OptionSources::IdParams.new(hg_cmd) }
+
+    it "skips param when set" do
+      hg_cmd.expects(:get_resource_id).with { |res| res.name != :domains }.at_least(0).returns(nil)
```
tstrachota: Do we actually need `.at_least(0)`? See my other comment.

test/unit/option_sources/ids_params_test.rb
mbacovsky: Sure, updated.

test/unit/option_sources/id_params_test.rb
mbacovsky: My thinking was that `at_least(0)` stands for `any times` which equivalent I didn't found in expectations docs. The thing is that the resource I use for testing in the command has other id params. When I set the expectation for the one param I care about I also need to cover the other calls to :get_resource_ids otherwise I get unexpected invocations. I used `at_least(0)` to indicate I don't care about the number of invocations. Is there better way to do this?

The other case has just one other param so the default `once()` works here. I've added at_least(0) to be consistent.

tstrachota:
Looks cajk to me, thanks @mbacovsky

PR #356, opened 2018-02-28
Fixes #22718 - raise error when wrong number of ids is resolved

mbacovsky:
reproducer, should fail with the patch:
```bash
hammer role create --name a
hammer user-group create --name ug1 --admin true --roles 'a,m'
```


mbacovsky:
@tstrachota do you think it is worth a stable branch update (0.12.1)?

tstrachota:
Together with http://projects.theforeman.org/issues/22658 I'd say yes.

tstrachota:
:+1: nice and backwards compatible. Thanks @mbacovsky !

PR #357, opened 2018-03-02
Fixes #22751 - temporary fix for override values

tstrachota:


tstrachota:
@ofedoren please test with 1.17 apidoc exports

mbacovsky:
@tstrachota, thanks! I was not able to come up with better solution. This works well and fixes the problem. I've run the test you changed against apidoc from 1.17 and it was green.
:+1:. I hope it will make it to 1.17 RC2 :fishing_pole_and_fish:

PR #359, opened 2018-03-16
Fixes #4609 - improve class import output

tstrachota:


test/functional/proxy_test.rb
```diff
@@ -0,0 +1,65 @@
...
+      }
+    end
+
+    it 'prints detailed report' do
+      api_expects(:smart_proxies, :import_puppetclasses, 'Import classes')
+        .with_params('id' => '83')
+        .returns(report)
+
+      output = OutputMatcher.new([
+        'Result:               Successfully updated environment and puppetclasses from the on-disk puppet installation',
```
mbacovsky: Can we do something about the gap between `Result:` and the message? I probably understand where it comes from but here it doesn't make much visual sense. Any idea?
```
Result:               Successfully updated environment and puppetclasses from the on-disk puppet installation
Changed environments: 
 1) production
    New classes: 
     1) apt::backports
     2) apt
     3) apt::params
```
Can we change the numbering for an unnumbered list? In my testing imports I had more then hundred new classes and the numbering does not improve readability.

tstrachota:
@mbacovsky I removed numbering and fixed the message for no changes. The only thing that we can currently do about the message indentation is to put it on a new line. Check if you like it more.

mbacovsky:
A way better now! Thanks @tstrachota. :poodle:  
```
Result:               
  Successfully updated environment and puppetclasses from the on-disk puppet installation
Changed environments: 
 1) production
    New classes: 
        elasticsearch::config
        elasticsearch
        elasticsearch::package
        elasticsearch::repo
```
It is good to go when the tests are green.

tstrachota:
@mbacovsky test failures are unrelated (http://projects.theforeman.org/issues/22964). Could you merge this?

mbacovsky:
[test]

mbacovsky:
:+1: This is good to go now! Thanks..

PR #360, opened 2018-03-22
Fixes #22964 - some numeric options are no longer recognized as numeric

tstrachota:


mbacovsky:
:+1: Looks good, thanks @tstrachota

PR #368, opened 2018-06-07
Fixes #20725 - retain data while host update in hammer

rahulbajaj0509:


tstrachota:
Thanks @rahulbajaj0509, makes sense to me. Could you please add some functional tests for your change?

rahulbajaj0509:
@tstrachota thanks for the input :) will do that asap!

lib/hammer_cli_foreman/hosts/common_update_options.rb
```diff
@@ -109,7 +109,6 @@ def domain_id(name)
       end
 
       def parameter_attributes
-        return [] unless option_parameters
```
tstrachota: I think that we still want to keep this to enable intentional wipe of parameters.

lib/hammer_cli_foreman/hosts/common_update_options.rb
```diff
@@ -120,8 +119,6 @@ def parameter_attributes
       end
 
       def nested_attributes(attrs)
-        return {} unless attrs
```
tstrachota: Ditto

lib/hammer_cli_foreman/hosts/common_update_options.rb
```diff
@@ -130,8 +127,6 @@ def nested_attributes(attrs)
       end
 
       def interfaces_attributes
-        return [] if option_interface_list.empty?
```
tstrachota: Ditto

junovitch:
Applying the changes in commit a649c05e4af7e3a3361aa9a24a56b39b0b8c717a look to resolve the issue I reported at https://community.theforeman.org/t/auto-deletion-of-hostparameters-on-foreman-1-15-causing-non-consistent-rebuilds/7291 and see in various bug trackers on the topic.

Thanks folks for your efforts tracking this down! Looking forward to seeing this finalized and hitting a release!

tstrachota:
Thanks for testing that @junovitch. It works as expected now @rahulbajaj0509, thanks for the changes.
The change still needs tests.

tstrachota:
**Just for the record:**
We discussed the issue with @rahulbajaj0509 on IRC. The current fix (using `unless option_parameters.nil?`) won't work because options that can be passed multiple times are never nil. See https://github.com/theforeman/hammer-cli/blob/master/lib/hammer_cli/options/option_definition.rb#L136-L138 for details. Using `[]` as a default value for multivalued options was a bad idea (shame on me, I'm the author of this very old code;)). I originally thought that removing that condition would help with fixing this issue but in fact, it would have no effect but making the code more readable (maybe). So it's not any dependency.

**I think that the correct fix is:**
Distinguishe between `create` and `update` actions and:
- for `create` pass the options as they are
- for `update` only send the options if they're not empty (because `[]` means no such option was entered)

test/functional/host_test.rb
```diff
@@ -87,6 +87,19 @@
     assert_cmd(expected_result, result)
   end
 
+  it "accepts host parameter attributes" do
+    params = ['--parameters', 'param1=value1']
+
+    api_expects(:hosts, :create, 'Create host with parameters') do |par|
+      par['host']['host_parameters_attributes'][0]['name'] == 'param1'
+      par['host']['host_parameters_attributes'][0]['value'] == 'value1'
```
tstrachota: BTW this code tests only the last line, you need to add `&&` at the end of the previous line.

rahulbajaj0509:
sorry for the delay on this one, @tstrachota i have tried to follow your comment, is this expected?

lib/hammer_cli_foreman/hosts/common_update_options.rb
```diff
@@ -76,10 +76,16 @@ def request_params
...
 
-        params['host']['host_parameters_attributes'] = parameter_attributes
+        params['host']['host_parameters_attributes'] = parameter_attributes unless option_parameters.nil?
         params['host']['compute_attributes'] = option_compute_attributes || {}
-        params['host']['compute_attributes']['volumes_attributes'] = nested_attributes(option_volume_list)
-        params['host']['interfaces_attributes'] = interfaces_attributes
+
+        if action == :update
+          params['host']['compute_attributes']['volumes_attributes'] = nested_attributes(option_volume_list) unless option_volume_list.empty?
+          params['host']['interfaces_attributes'] = interfaces_attributes unless option_interface_list unless option_interface_list.empty?
```
tstrachota: There's double `unless`.
`option_interface_list` is going to be always true, we should only keep `option_interface_list.empty?`

lib/hammer_cli_foreman/hosts/common_update_options.rb
```diff
@@ -76,10 +76,16 @@ def request_params
...
+        else
+          params['host']['compute_attributes']['volumes_attributes'] = nested_attributes(option_volume_list)
+          params['host']['interfaces_attributes'] = interfaces_attributes unless option_interface_list
```
tstrachota: As I said in my previous comment, `option_interface_list` is always true and that makes this condition to be always false -> the attributes will be nil. The tests are actually failing for that right now.

iNecas:
@tstrachota mind having another look? Seems like your comments might be already addressed and the tests are passing now.

@rahulbajaj0509 it's not a bad idea to let the reviewer know when you've addressed some comments: AFAIK github doesn't provide notification when there is a new commit in the PR

tstrachota:
Thanks for the notification @iNecas. Github really doesn't send any notification for new commits so it's a good idea to leave a comment upon your changes indeed @rahulbajaj0509.

All looks good now and it works fine, merging. Thanks @rahulbajaj0509 !

PR #369, opened 2018-06-20
Fixes #23722 - removed redundant info about OS from hostgroup

kgaikwad:


tstrachota:
Works as expected, thank you @kgaikwad !

PR #373, opened 2018-07-09
Fixes #15674 - add disassociate command to host

shiramax:


lib/hammer_cli_foreman/host.rb
```diff
@@ -428,6 +428,17 @@ def print_data(data)
       build_options
     end
 
+    class DisassociateCommand < HammerCLIForeman::SingleResourceCommand
+      action :disassociate
+
+      command_name "disassociate"
+      desc _("Disassociate a host")
+      success_message _("The host has been disassociated from VM")
```
tstrachota: Please add error message too.

test/unit/host_test.rb
```diff
@@ -416,4 +416,13 @@
 
     end
   end
+
+
+  context "DisassociateCommand" do
```
tstrachota: Would you mind adding also a functional test for the list command? We prefer covering commands with functional tests because they provide much more readable errors, test the command in the context of the whole cli and uses real data (compared to default responses from apidoc).
You can see an example here: https://github.com/theforeman/hammer-cli-foreman/blob/master/test/functional/personal_access_token_test.rb#L55-L80

shiramax:
@tstrachota done, please review again :)

test/functional/host_test.rb
```diff
@@ -219,3 +219,29 @@
     result.exit_code.must_equal HammerCLI::EX_OK
   end
 end
+describe 'disassociate host from vm' do
+  let(:cmd) { ["host","disassociate"] }
```
tstrachota: Nitpick: space after `,` please

test/functional/host_test.rb
```diff
@@ -219,3 +219,29 @@
...
+
+  it 'successful disassociate host' do
+    params = ['--id=1']
+    expected_result = success_result("The host has been disassociated from VM\n")
+    api_expects(:hosts, :disassociate, 'disassociate hosts') do |params|
+      params['id'] == "1"
+    end.returns()
```
tstrachota: Please give it some return value. `returns({})` should be enough in this case. It will make the return value more obvious. I'm not sure what's the behavior of `returns` without any value.

test/functional/host_test.rb
```diff
@@ -219,3 +219,29 @@
...
+
+    result = run_cmd(cmd + params)
+    assert_cmd(expected_result, result)
+  end
+
+  it 'prints error on missing host id' do
+    params = []
```
tstrachota: Up to you but we can probably leave this off.

test/functional/host_test.rb
```diff
@@ -219,3 +219,29 @@
...
+    expected_result = CommandExpectation.new
+    expected_result.expected_err = [
+      "Failed to disassociated host from VM:",
+      "  Missing arguments for 'id'", ''].join("\n")
```
tstrachota: Nitpick: Could you please put `''` on a new line so that it's more obvious we're expecting one more new line here?

shiramax:
@tstrachota thanks :) 
could you please review again?

tstrachota:
:+1: thanks you @shiramax 
Btw this feature is a good candidate for the next community demo.

PR #374, opened 2018-07-10
Fixes #14510 - Hammer uses /config_reports rather than /reports

ofedoren:


tstrachota:
Thanks @ofedoren. It all looks sane to me but I need to test it. If there's any volunteer who could do the testing, I'd appreciate it.

mbacovsky:
@tstrachota I'll do the testing

tstrachota:
Thanks @mbacovsky Feel free to merge if it works fine.

lib/hammer_cli_foreman/host.rb
```diff
@@ -213,15 +213,14 @@ def send_request
       end
     end
 
+    class ConfigReportsCommand < HammerCLIForeman::ListCommand
+      command_name 'config-reports'
```
tstrachota: Hmm, we should probably keep the `reports` command with a deprecation warning for the sake of compatibility.

lib/hammer_cli_foreman/config_report.rb
```diff
@@ -1,8 +1,8 @@
 module HammerCLIForeman
 
-  class Report < HammerCLIForeman::Command
+  class ConfigReport < HammerCLIForeman::Command
```
mbacovsky: @ofedoren, there is still reference to `Report` in hammer_cli_foreman.rb L105. The "global" report command there should be also renamed. 

I agree with Tomas's point about keeping the report command as deprecated for a version or two.

ofedoren:
@mbacovsky, @tstrachota thanks!

Updated: Both `report` and `host reports` commands have deprecation warning now.

mbacovsky:
:+1: Well done, thanks @ofedoren!

PR #375, opened 2018-07-10
Fixes #12189 - hammer report info doesn't show logs resources and mes…

bjoernneumann:
Source and Message will show something now when shown with hammer report info --id=<someid> instead of
 1) Resource:
    Message:

### DIFF
--- a/lib/hammer_cli_foreman/config_report.rb
+++ b/lib/hammer_cli_foreman/config_report.rb
@@ -57,13 +57,11 @@ module HammerCLIForeman
           end
         end
         field :logs, _("Logs"), Fields::Collection do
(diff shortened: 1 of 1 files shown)


mbacovsky:
@bjoernneumann, thank you for the contribution, it is appreciated!

Recently the `report.rb` was renamed to `config_report.rb` to match the changes in the API. Your patch looks good to be merged,  would you mind to move it to the new location so that it could be tested and merged?

bjoernneumann:
@mbacovsky Done

mbacovsky:
:+1: @bjoernneumann, perfect! Thanks for the update. Merging.

PR #377, opened 2018-07-18
Fixes #24285 - propagate the error message from core

ares:


ares:
needs https://github.com/theforeman/foreman/pull/5848 to be applied on core first, without the patch, we'd start seeing `Unable to authenticate user admin` only instead of `Invalid username or password` which is not truth for each authentication issue.

```sh
ares@tony:[~/z/hammer][master✗]$ be bin/hammer --username admin --password xchangeme user list 
Unable to authenticate user admin
ares@tony:[~/z/hammer][master✗]$ be bin/hammer --username admin --password xchangeme user list
Too many tries, please try again in a few minutes.
ares@tony:[~/z/hammer][master✗]$ be bin/hammer --username admin --password xchangeme user list
Unable to authenticate user admin
```

tstrachota:
Works fine, thanks @ares !

PR #380, opened 2018-07-31
Fixes #24490 - stop overriding apipie help for host flags

ares:


lib/hammer_cli_foreman/hosts/common_update_options.rb
```diff
@@ -25,12 +25,6 @@ def self.included(base)
         bme_options = {}
         bme_options[:default] = 'true' if base.action.to_sym == :create
 
-        bme_options[:format] = HammerCLI::Options::Normalizers::Bool.new
-        base.option "--managed", "MANAGED", " ", bme_options
-        bme_options[:format] = HammerCLI::Options::Normalizers::Bool.new
-        base.option "--build", "BUILD", " ", bme_options
```
mbacovsky: I agree that this change makes sense and fixes the help strings. But it also removes the default True values. While API defaults to False for build option this patch changes the behavior of the `host create` command.

As a solution you could set the default values for these options and create action in the `request_params` below.

test/unit/host_test.rb
```diff
@@ -233,9 +233,6 @@
             "--ip=1.2.3.4", "--mac=11:22:33:44:55:66", "--medium-id=1", "--partition-table-id=1", "--subnet-id=1",
             "--model-id=1", "--hostgroup-id=1", "--owner-id=1", '--puppet-ca-proxy-id=1', '--puppet-class-ids',
             "--root-password=pwd", "--ask-root-password=true", "--provision-method=build", "--interface=primary=true,provision=true"] do
-        it_should_call_action_and_test_params(:create) { |par| par["host"]["managed"] == true }
```
mbacovsky: We should keep the tests to make sure we don't change the behavior

ares:
updated, please take another look

mbacovsky:
:+1: Works as expected. Thanks @ares!

PR #382, opened 2018-08-16
Fixes #20609 Subnet create/update should accept proxy names

apuntamb:


test/functional/subnet/create_test.rb
```diff
@@ -0,0 +1,184 @@
...
+      end
+
+      it 'allows dhcp name' do
+        api_expects(:smart_proxies, :index) do |p|
+          p[:search] = "name = \"sp1\""
+        end.returns(index_response([{'id' => 1}]))
+        api_expects(:subnets, :create) do |p|
+          p['subnet']['dhcp_id'] == 1 &&
+            p['subnet']['name'] == 'net1'
+        end
```
tstrachota: Not that important but just for you to know, you can use `with_params` here:
```ruby
api_expects(:subnets, :create).with_params({
  :subnet => {
    :name => 'net1',
    :dhcp_id => 1
  }
})
```
It performs the same test but the benefit is much more readable error message if it fails.
Compare
```
expected params to match block at: /home/vagrant/hammer/hammer-cli-foreman/test/functional/subnet/create_test.rb:31
```
with output produced by `with_params`
```
  expected params to include: {
    "subnet": {
      "name": "net12",
      "dhcp_id": 1
    }
  }
```

tstrachota: One more idea I got (feel free to ignore it if you don't like it) is defining a `subnet_params` helper somewhere up in the test:
```ruby
def subnet_params(additional_params = {})
  params = {
    :subnet => {
      :name => 'net1'
    }
  }
  params[:subnet].merge!(additional_params)
  params
end
```
That would reduce code repetition and enable writing shorter tests like:
```ruby
api_expects(:subnets, :create).with_params(subnet_params(:tftp_id => '1'))
```

But your solution works and I'm fine with so I'm leaving it up to you to consider whether it makes sense to add the helper.

lib/hammer_cli_foreman/subnet.rb
```diff
@@ -1,5 +1,28 @@
 module HammerCLIForeman
 
+   module SubnetUpdateCreateCommons
+
+    def self.included(base)
+      base.option "--dns", "DNS_NAME", _("DNS Proxy Name to use within this subnet")
```
tstrachota: Maybe just "DNS Proxy to use within this subnet" to match the strings in UI. Similarly for the other strings.

test/functional/subnet/create_test.rb
```diff
@@ -0,0 +1,184 @@
...
+            :name => 'net1',
+            :dhcp_id => '1'
+          }
+        })
+        run_cmd(%w(subnet create --name net1 --network=192.168.122.0 --mask=255.255.255.0 --dhcp-id 1))
+      end
+
+      it 'allows dhcp name' do
+        api_expects(:smart_proxies, :index) do |p|
+          p[:search] = "name = \"sp1\""
```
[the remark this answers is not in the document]
apuntamb: Thanks for the review @tstrachota . I liked the idea of using api_expects_search and hence, will be pushing new changes with the same.

test/functional/subnet/create_test.rb
apuntamb: Thanks for suggesting the helper method. I would definitely like to consider adding the helper and accordingly writing the tests.

PR #385, opened 2018-09-10
Fixes #20187 - Display Host Status in Hammer

akshay-mendhe:


lib/hammer_cli_foreman/host.rb
```diff
@@ -66,6 +67,8 @@ def extend_data(host)
         field nil, _("Puppet Master Proxy"), Fields::SingleReference, :key => :puppet_proxy
         field :certname, _("Cert name")
         field :managed, _("Managed"), Fields::Boolean
+        field :build_status_label, _("Build Status")
+        field :global_status_label, _("Global Status")
```
tstrachota: Would you mind moving the global status above the build one? I'd say it makes more sense to see it on the first place.

lib/hammer_cli_foreman/host.rb
```diff
@@ -66,6 +67,8 @@ def extend_data(host)
         field nil, _("Puppet Master Proxy"), Fields::SingleReference, :key => :puppet_proxy
         field :certname, _("Cert name")
         field :managed, _("Managed"), Fields::Boolean
+        field :build_status_label, _("Build Status")
```
tstrachota: Would it make sense to encapsulate both statuses with `label _('Status') do`?
It would probably make it more readable to have all statuses in one section. Also it would be easier for plugins to add their own statuses. That should be soon enabled by https://github.com/theforeman/hammer-cli/pull/275.

akshay-mendhe:
@tstrachota Updated the changes.

tstrachota:
Thank you @akshay-mendhe. This change is a good candidate for the upcoming community demo!

akshay-mendhe:
@tstrachota Thank You!

PR #386, opened 2018-09-13
Fixes #3651 - Provides a way of listing compute attribute values

ofedoren:


tstrachota:
@ofedoren this looks very good! Would you mind using functional tests instead of the unit ones? It's a preferred way of testing the commands with full api interactions.

ofedoren:
@tstrachota sure. So should I remove the unit tests I added? After the functional tests are added of course.

tstrachota:
Yeah, it makes sense to me to keep only the functional ones in this case.

ofedoren:
@tstrachota, updated.

I added the functional tests and also changed the order of `Id` and `Name` fields, so it looks like the other `List` commands.

mbacovsky:
@tstrachota, @ofedoren: Do we want to polish the NotImplemented errors or is it not a problem?
```
# hammer compute-resource clusters --id 1                                     
ERF42-7329 [Foreman::Exception]: Not implemented for Libvirt
```

tstrachota:
Hmm, it's not that big deal for me but removing the "header" `ERF42-7329 [Foreman::Exception]:` would probably make it nicer. If it's too complicated to implement, I'm ok with leaving it as is.

tstrachota:
Btw @ofedoren this is a great addition that many people will appreciate. It's a very nice candidate for the next community demo.

PR #388, opened 2018-09-24
Fixes #21764 - Add Location and Description Fields

akshay-mendhe:
Adds Location and Description Field for Host Command

### DIFF
--- a/lib/hammer_cli_foreman/host.rb
+++ b/lib/hammer_cli_foreman/host.rb
@@ -34,6 +34,8 @@ module HammerCLIForeman
         field :ip, _("IP")
         field :mac, _("MAC")
         field :global_status_label, _("Global Status")
+        field nil, _("Location"), Fields::SingleReference, :key => :location, :hide_blank => true
+        field :comment, _("Additional Information")
       end
 
       build_options :without => [:include, :thin]
--- a/test/unit/host_test.rb
+++ b/test/unit/host_test.rb
@@ -23,7 +23,7 @@ describe HammerCLIForeman::Host do
       let(:expected_record_count) { count_records(cmd.resource.call(:index)) }
 
       it_should_print_n_records
-      it_should_print_columns ["Id", "Name", "Operating System", "Host Group", "IP", "MAC", "Global Status"]
+      it_should_print_columns ["Id", "Name", "Operating System", "Host Group", "IP", "MAC", "Global Status", "Location"]
     end
 
   end


akshay-mendhe:
Added Location field, but what should be there in Description field?

akshay-mendhe:
We can show column "additional information" about host if there is any additional information given while creating host.

mbacovsky:
@akshay-mendhe thank you for the contribution and sorry for not responding for so long. 
The code looks good and I appreciate the test updated. Adding new fields to `host list` is quite sensitive task as it may make the list too wide and difficult to use. 

The long term plan is to make the fields configurable to better match the users needs. While the feature is being worked on (https://github.com/theforeman/hammer-cli/pull/276) currently I'd prefer to wait and update this PR to leverage the new feature when it is merged. I'll update here when that happens. Would that be acceptable?

akshay-mendhe:
Yes, sure Thanks @mbacovsky

mbacovsky:
@chris1984 as I wrote above, extending the host list width is quite sensitive. Especially adding the description will lead to line wraps in the middle and unreadable output. I hope we will merge https://github.com/theforeman/hammer-cli/pull/276 next week and unblock this with new possibilities.

ofedoren:
@mbacovsky, https://github.com/theforeman/hammer-cli/pull/276 is merged, what was your suggestion?

mbacovsky:
My suggestion now is to add the new fields only to ALL set so that the fields are not present in the DEFAULT set:
```ruby7
field nil, _("Location"), Fields::SingleReference, :key => :location, :hide_blank => true, :sets => ['ALL']
field :comment, _("Additional Information"), nil, :sets => ['ALL']
```
then
`hammer host list` won't print them while `hammer host list --fields ALL` will do.
Also the original issue speaks about the organization too so it can be added as well.

mbacovsky:
Note the `--fields` feature require https://github.com/theforeman/hammer-cli/pull/307 to work as expected.

mbacovsky:
@akshay-mendhe could you please rebase and check my previous comments? Would it work for you?

kgaikwad:
@mbacovsky, 
It would be good if we create new pull-request with suggested changes and close this one on behalf of new.  Would it be possible for you? or let me know I will create one.

mbacovsky:
@kgaikwad if that works for you I'm okay with that.

kgaikwad:
@mbacovsky, 
Can we close this pull-request on behalf of [PR-423](https://github.com/theforeman/hammer-cli-foreman/pull/423)?

mbacovsky:
@kgaikwad thanks for the new PR! Closing this one.

PR #391, opened 2018-10-03
Fixes #14832 - Request password regardless of redirection

ofedoren:
Requires https://github.com/theforeman/hammer-cli/pull/289

### DIFF
--- a/lib/hammer_cli_foreman/api/interactive_basic_auth.rb
+++ b/lib/hammer_cli_foreman/api/interactive_basic_auth.rb
@@ -1,5 +1,3 @@
-require 'highline/import'
-
 module HammerCLIForeman
   module Api
     class InteractiveBasicAuth < ApipieBindings::Authenticators::BasicAuth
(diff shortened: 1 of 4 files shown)


mbacovsky:
@ofedoren, thanks for the patches! It works fine. I think we should convert also the other places where we use `ask`:
- https://github.com/theforeman/hammer-cli-foreman/blob/1388c96fae176950ed8c9bfcd82e137f0079b88c/lib/hammer_cli_foreman/hostgroup.rb#L24
- https://github.com/theforeman/hammer-cli-foreman/blob/cef2d066ead6789dd08656466c8d614463b948b1/lib/hammer_cli_foreman/option_sources/user_params.rb#L46
- https://github.com/theforeman/hammer-cli-foreman/blob/1b8d8237d55661e2c956b3a6052688367bfee559/lib/hammer_cli_foreman/hosts/common_update_options.rb#L50

Those are all I've found in our codebase

ofedoren:
Updated.

@mbacovsky thanks a lot! Somehow I've missed those parts.

mbacovsky:
:+1: Looks good now! Thanks @ofedoren.

PR #397, opened 2018-11-21
Fixes #11431 - Include information about search fields

ofedoren:
requires [Extensible commands](https://github.com/theforeman/hammer-cli/pull/291).


tstrachota:
Nice improvement @ofedoren. I'd prefer if the parent list command would detect that metadata are present and add the "Search fields" help section automatically without the need to call `extend_with` on each and every list command. To me this is more important than using plugin extensions here.

ofedoren:
@tstrachota, thanks! I agree that the previous approach was not quite good. Is that better now?

ofedoren:
@tstrachota, updated. Thanks for your suggestions! I thought about `One of: <values>` instead of `Values: <values>`, but I think this is good enough as well.

PR #398, opened 2018-11-22
BZ #20538 - add compute profile commands

shiramax:


shiramax:
@mbacovsky I created a new PR for compute profile as you requested.
Thanks :)

lib/hammer_cli_foreman/compute_profile.rb
```diff
@@ -0,0 +1,83 @@
...
+    class DeleteCommand < HammerCLIForeman::DeleteCommand
+      success_message _('Compute profile deleted.')
+      failure_message _('Could not delete the Compute profile')
+      validate_options do
+        any(:option_name,:option_id).required
+      end
+      build_options
+    end
+
+    lazy_subcommand('attribute', _("Create update and delete Compute Resource attributes"),
```
mbacovsky: According to our discussion `attribute` should be renamed to `values` with the description "Create, update and delete compute attributes, interfaces and volumes"

lib/hammer_cli_foreman/compute_resource/vmware.rb
```diff
@@ -0,0 +1,58 @@
...
+      private
+
+      def interface_type_description(h)
+        [
+            _('Type of the network adapter, for example one of:'),
+            h.indent(INTERFACE_TYPES),
+            _("See documentation center for your version of vSphere to find more details about available adapter types:"),
+            h.indent('https://www.vmware.com/support/pubs/')
+        ].join("\n")
+      end
```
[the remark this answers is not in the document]
mbacovsky: @shiramax as it is just an item description as everything else I think we can sacrifice the `h.indent`s and make it plain text (untested):
```ruby
[
  _('Type of the network adapter, for example one of:'),
￼  INTERFACE_TYPES.map { |it| '  ' + it },
￼  _("See documentation center for your version of vSphere to find more details about available adapter types:"),
￼  '  https://www.vmware.com/support/pubs/'
￼].flatten(1).join("\n")
```
The harder way might be to use markdown and implement conversion from markdown to helpbuilder which may be interesting component to the extended help we discussed earlier. It is probably out of the scope of this PR though.

lib/hammer_cli_foreman/compute_resources.rb
```diff
@@ -0,0 +1,43 @@
+require '../hammer-cli-foreman/lib/hammer_cli_foreman/compute_resource/base'
+Dir['../hammer-cli-foreman/lib/hammer_cli_foreman/compute_resource/*'].each {|file| require file }
+
+module HammerCLIForeman
+  module ComputeResourcesNew
```
mbacovsky: This module can be removed

lib/hammer_cli_foreman/compute_resources.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  end
+  def self.compute_resources
+    @compute_resources = {
+        'ec2' => HammerCLIForeman::ComputeResources::EC2.new,
```
mbacovsky: This way each time you call `HammerCLIForeman.compute_resources` all the resources are re-created.
It would be better if there is HammerCLIForeman.register_compute_resource(name, compute_resource) which is called from each of the compute resource files to register the new resource. Another benefit of this is that if you add new resource from e.g. plugin, you don't need to modify this method.

mbacovsky:
@shiramax thanks for this update. I like how you reworked the compute resources. It looks much cleaner now. there are some minor comments inline.

I played a bit with the --interface and --volume params and I think we should make it the same way as in host create/update commands. So to keep `--interface` and `--volume` and keep them multivalued. I think it does not make sense to keep the `--volume-id` and `--interface-id` because it changes how the `--interface` parameter work and is confusing.

It seems we are almost there, could you please rebase?

shiramax:
@mbacovsky thanks, please review again

mbacovsky:
@shiramax the value create and update options looks good now. I'll do some more testing. 
It seems you've missed my comments about registering of the compute resources from the resource definition. Is the comment and reasoning clear?
Could you please add the updates in separate commits? It is much easier for me to review the changes. Thanks.

lib/hammer_cli_foreman/attribute.rb
```diff
@@ -0,0 +1,282 @@
+require 'hammer_cli_foreman/image'
+require 'hammer_cli_foreman/compute_resources'
+
+module HammerCLIForeman
+  class Attribute < HammerCLIForeman::Command
```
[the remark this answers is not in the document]
mbacovsky: *when going through the files

shiramax:
@mbacovsky, hey thanks for the review.
 Sorry for missing your comment about registering the key, I thought that I can fix the problem with recreating @compute_resources hash each time in a different way. 
But I will try to do the register part, the problem is that i'm not sure where the function register_compute_resource should be, should it be in hammer-cli-foreman/lib/hammer_cli_foreman/compute_resources.rb, if i think it will cause problems in the required files.
thanks.

lib/hammer_cli_foreman/attribute.rb
```diff
@@ -0,0 +1,282 @@
...
+      Time.now.to_i.to_s
+    end
+
+    def self.get_params(options)
+      params = {}
+      params['compute_attribute'] = {}
+      profile = HammerCLIForeman.record_to_common_format(
+          HammerCLIForeman.foreman_resource(:compute_profiles).call(:show, 'id' => options['option_compute_profile_id'] )
+      )
+      HammerCLIForeman.foreman_resource(:compute_resources).call(:show, 'id' => options['option_compute_resource_id'] )
```
mbacovsky: What is purpose of this extra call? The values does not seem to be stored/used

lib/hammer_cli_foreman/attribute.rb
```diff
@@ -0,0 +1,282 @@
...
+      params['compute_attribute'] = profile['compute_attributes'].select { |hash| hash['compute_resource_id'] == options['option_compute_resource_id']}[0] || {}
+      params
```
mbacovsky: I think we need to remove the 'attributes' which is read-only (?) copy of the compute_attributes. 
```ruby
params["compute_attribute"].delete("attributes")
```

lib/hammer_cli_foreman/attribute.rb
```diff
@@ -0,0 +1,329 @@
...
+
+      validate_options do
+        any(:option_compute_profile_id, :option_compute_profile_name ).required
+        any(:option_compute_resource_id, :option_compute_resource_name).required
+      end
+
+      def request_params
+        params = super
+        params['compute_attribute']['vm_attrs'] = option_compute_attributes || {}
+        if option_interface
```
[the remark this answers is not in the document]
mbacovsky: This is still valid. `option_interface` is now `option_interface_list` and needs different handling. See my comment in Update command

lib/hammer_cli_foreman/attribute.rb
```diff
@@ -0,0 +1,282 @@
...
+             :format => HammerCLI::Options::Normalizers::KeyValueList.new, :multivalued => true
+
+      extend_help do |h|
+        ::HammerCLIForeman::ComputeResources.extend_help(h, :all)
+      end
+
+      validate_options do
+        any(:option_compute_profile_id, :option_compute_profile_name ).required
+        any(:option_compute_resource_id, :option_compute_resource_name).required
+        any(:option_interface, :option_volume, :option_compute_attributes).required
```
mbacovsky: This validation won't work. Our validators do not work well with multivalued options. We either need to fix the validator or remove this validation.

lib/hammer_cli_foreman/attribute.rb
```diff
@@ -0,0 +1,282 @@
...
+      end
+
+      def request_params
+        params = HammerCLIForeman::Attribute.get_params(options)
```
mbacovsky: @shiramax, while testing the update it didn't work as expected. E.g. 
```
hammer  compute-profile values update --compute-profile-id 1 --compute-resource-id 1 --compute-attributes cpus=8 --volume capacity=8G
``` 
after fixing some issues caused I've lost interfaces and volumes attributes. The `option_compute_attributes` replaced all `vm_attrs` including the volumes and interfaces. The volumes then failed as it was sent as a list. I'd expect the example command will replace compute_attributes, volume_attributes and keep existing interface attributes. I also found out the 'interfaces_attributes` should be called `nics_attributes`. Not sure if both are accepted but nics_attributes worked for me.

Overall patch to the `request_params` that worked for me would be:
```diff
diff --git a/lib/hammer_cli_foreman/attribute.rb b/lib/hammer_cli_foreman/attribute.rb
index 0fd4e60..525cecf 100644
--- a/lib/hammer_cli_foreman/attribute.rb
+++ b/lib/hammer_cli_foreman/attribute.rb
@@ -15,8 +15,8 @@ module HammerCLIForeman
       profile = HammerCLIForeman.record_to_common_format(
           HammerCLIForeman.foreman_resource(:compute_profiles).call(:show, 'id' => options['option_compute_profile_id'] )
       )
-      HammerCLIForeman.foreman_resource(:compute_resources).call(:show, 'id' => options['option_compute_resource_id'] )
       params['compute_attribute'] = profile['compute_attributes'].select { |hash| hash['compute_resource_id'] == options['option_compute_resource_id']}[0] || {}
+      params["compute_attribute"].delete("attributes")
       params
     end
 
@@ -73,22 +73,35 @@ module HammerCLIForeman
       validate_options do
         any(:option_compute_profile_id, :option_compute_profile_name ).required
         any(:option_compute_resource_id, :option_compute_resource_name).required
-        any(:option_interface, :option_volume, :option_compute_attributes).required
       end
 
       def request_params
         params = HammerCLIForeman::Attribute.get_params(options)
+        vm_attrs = params['compute_attribute']['vm_attrs']
+        original_volumes = vm_attrs['volumes_attributes'] || {}
+        original_interfaces = vm_attrs['nics_attributes'] || {}
         params['id'] =  params['compute_attribute']['id']
-        params['compute_attribute']['vm_attrs'] = option_compute_attributes if option_compute_attributes
-        params['compute_attribute']['vm_attrs']['interfaces_attributes'] = option_interface if option_interface
-        params['compute_attribute']['vm_attrs']['volumes_attributes'] = option_compute_attributes if option_volume
+        if option_compute_attributes
+          vm_attrs = option_compute_attributes
+          vm_attrs['volumes_attributes'] ||= original_volumes
+          vm_attrs['nics_attributes'] ||= original_interfaces
+        end
+        vm_attrs['nics_attributes'] = attribute_hash(option_interface_list) unless option_interface_list.empty?
+        vm_attrs['volumes_attributes'] = attribute_hash(option_volume_list) unless option_volume_list.empty?
+        params['compute_attribute']['vm_attrs'] = vm_attrs
         params
       end
+
+      def attribute_hash(attribute_list)
+        attribute_list.size.times.map { |idx| idx.to_s }.zip(attribute_list).to_h
+      end
+
       success_message _('Compute profile attributes updated.')
       failure_message _('Could not update the compute profile attributes')
 
       build_options :without => :id
     end
+
     # Using the Update command because adding a new interface is done by modifying existing compute_attribute
     class AddInterface < HammerCLIForeman::UpdateCommand
       command_name 'add-interface'
```
Could you check if it matches your expectations and confirm it works for you?
If so I think we will need to update also the other create/update commands. Let me know if you need some help.

lib/hammer_cli_foreman/attribute.rb
```diff
@@ -0,0 +1,303 @@
...
+      end
+      success_message _('Compute profile attributes updated.')
+      failure_message _('Could not update the compute profile attributes')
+
+      build_options :without => :id
+    end
+
+    # Using the Update command because adding a new interface is done by modifying existing compute_attribute
+    class AddInterface < HammerCLIForeman::UpdateCommand
+      command_name 'add-interface'
```
pondrejk: somehow when doing `hammer compute-profile values add-interface -h`, provider specific options don't get populated:
```
 hammer compute-profile values add-interface -h
Usage:
    hammer compute-profile values add-interface [OPTIONS]

Options:
 --compute-profile COMPUTE_PROFILE_NAME    Compute profile name
 --compute-profile-id COMPUTE_PROFILE_ID    
 --compute-resource COMPUTE_RESOURCE_NAME  Compute resource name
 --compute-resource-id COMPUTE_RESOURCE_ID  
 --interface SET_VALUES                    Interface parameters, should be comma separated list of values
                                           Comma-separated list of key=value
 --location LOCATION_NAME                  Location name
 --location-id LOCATION_ID                  
 --location-title LOCATION_TITLE           Location title
 --organization ORGANIZATION_NAME          Organization name
 --organization-id ORGANIZATION_ID         Organization ID
 --organization-title ORGANIZATION_TITLE   Organization title
 -h, --help                                Print help

Provider specific options:
  ec2:


  gce:


  libvirt:


  openstack:


  ovirt:


  rackspace:


  vmware:
```
This happens only for add-interface other values subcommand have provider specific help as expected

mbacovsky:
[test hammer]

test/functional/attribute_test.rb
```diff
@@ -0,0 +1,621 @@
...
+
+    it "should print error on missing --compute-resource-id or --compute-resource" do
+      params = ['--compute-profile=A', '--interface-id=1']
+      expected_result = usage_error_result(
+        @cmd,
+        "At least one of options --compute-resource-id, --compute-resource is required.",
+        "Could not remove interface"
+      )
+
+      api_expects_no_call
```
mbacovsky: This test is failing as there is '--compute-profile=A' in the params which triggers name-resolution API call. Either change the param to '--compute-profile-id=1' (preferred) or add API call expectation to the test. 
The same pattern is causing failure in some other tests too.

lib/hammer_cli_foreman/attribute.rb
```diff
@@ -0,0 +1,303 @@
...
+      params['compute_attribute'].delete('attributes') if params['compute_attribute']['attributes']
+      params
+    end
+
+    def self.attribute_hash(attribute_list)
+      attribute_list.size.times.map { |idx| idx.to_s }.zip(attribute_list).to_h
+    end
+
+    class Create < HammerCLIForeman::CreateCommand
+        desc _('Create compute profile set of values.')
```
mbacovsky: The command descriptions shouldn't end with '.'. Please see the test results for other occurrences.

shiramax: done

test/functional/attribute_test.rb
shiramax: done

lib/hammer_cli_foreman/attribute.rb
shiramax: done

lib/hammer_cli_foreman/compute_resource/register_compute_resources.rb
```diff
@@ -0,0 +1,18 @@
...
+    @compute_resources[name] = compute_resource
+  end
+  require '../hammer-cli-foreman/lib/hammer_cli_foreman/compute_resource/base'
+  require '../hammer-cli-foreman/lib/hammer_cli_foreman/compute_resource/ec2.rb'
+  require '../hammer-cli-foreman/lib/hammer_cli_foreman/compute_resource/gce.rb'
+  require '../hammer-cli-foreman/lib/hammer_cli_foreman/compute_resource/libvirt.rb'
+  require '../hammer-cli-foreman/lib/hammer_cli_foreman/compute_resource/openstack.rb'
+  require '../hammer-cli-foreman/lib/hammer_cli_foreman/compute_resource/ovirt.rb'
+  require '../hammer-cli-foreman/lib/hammer_cli_foreman/compute_resource/rackspace.rb'
+  require '../hammer-cli-foreman/lib/hammer_cli_foreman/compute_resource/vmware.rb'
```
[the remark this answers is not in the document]
shiramax: done

lib/hammer_cli_foreman/compute_resource/ovirt.rb
```diff
@@ -0,0 +1,36 @@
+module HammerCLIForeman
+  module ComputeResources
+    class Ovirt < Base
+      def name
+        _('oVirt')
+      end
+
```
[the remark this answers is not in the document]
shiramax: thanks. 
I fixed that issue.
 apparently, there are different names to the interfaces_attributes in different provider.

shiramax:
@mbacovsky please review again :)

lib/hammer_cli_foreman/compute_resource/libvirt.rb
```diff
@@ -15,9 +15,9 @@ def compute_attributes
 
       def interface_attributes
         [
-            ['compute_type',                     _('Possible values: %s') % 'bridge, network'],
-            ['compute_network / compute_bridge', _('Name of interface according to type')],
-            ['compute_model',                    _('Possible values: %s') % 'virtio, rtl8139, ne2k_pci, pcnet, e1000']
+            ['type',                     _('Possible values: %s') % 'bridge, network'],
```
mbacovsky: Pleas fix the indentation

lib/hammer_cli_foreman/attribute.rb
```diff
@@ -42,8 +42,14 @@ class Create < HammerCLIForeman::CreateCommand
 
       def request_params
         params = super
+        compute_resource_name = HammerCLIForeman.record_to_common_format(
+            HammerCLIForeman.foreman_resource(:compute_resources).call(:show, 'id' => option_compute_resource_id )
+        )["name"]
+        interfaces_attrs_name_list = {"libvirt" => "nics_attributes","gce" => "network_interfaces_nics_attributes"}
```
mbacovsky: This info is feature of compute resource, shouldn't it be in the appropriate class?
e.g. `HammerCLIForeman.compute_resources[compute_resource_name].try('interfaces_attr_name')`

lib/hammer_cli_foreman/compute_resource/help_utils.rb
```diff
@@ -1,17 +1,24 @@
 module HammerCLIForeman
   module ComputeResources
-    def self.extend_help(h, attributes)
+    def self.extend_help(h, attributes, add_host_specific_attrs = false )
```
mbacovsky: Could you use named attribute for more readable method signature?
```ruby
def self.extend_help(h, attributes, add_host_specific_attrs: false)
end

::HammerCLIForeman::ComputeResources.extend_help(h, :all, add_host_specific_attrs: true)
```

lib/hammer_cli_foreman/compute_resource/libvirt.rb
```diff
@@ -15,9 +15,9 @@ def compute_attributes
...
+            ['bridge', _('Name of interface according to type')],
+            ['model',                    _('Possible values: %s') % 'virtio, rtl8139, ne2k_pci, pcnet, e1000']
```
mbacovsky: indentation

lib/hammer_cli_foreman/attribute.rb
mbacovsky: It could set in the `Base` to "interfaces_attributes" and over-ridden in resources where needed. As it is now there is no way to easily add new resource with non-standard nics naming

lib/hammer_cli_foreman/compute_resource/ovirt.rb
```diff
@@ -37,6 +37,6 @@ def volume_attributes;
         ]
       end
     end
-    HammerCLIForeman.register_compute_resource('ovirt', Ovirt.new)
+    HammerCLIForeman.register_compute_resource('oVirt', Ovirt.new)
```
mbacovsky: As it is not sometimes very clear how the compute resources are named we should use the names in lowercase as it was originally.

lib/hammer_cli_foreman/compute_resource/register_compute_resources.rb
```diff
@@ -1,11 +1,17 @@
 module HammerCLIForeman
   @compute_resources = {}
+  @interfaces_attrs_name_list ={}
```
mbacovsky: What benefit it has to store this duplicate information in here? I'd prefer to not add to top lavel namespace more things then it is necessary.
```ruby
HammerCLIForeman.compute_resources['ovirt'].interfaces_attrs_name
```
feels readable enough.

test/functional/attribute_test.rb
```diff
@@ -0,0 +1,654 @@
+require File.join(File.dirname(__FILE__), 'test_helper')
```
mbacovsky: This file should be renamed to `compute_attribute_test.rb` to match the module file it is testing.

lib/hammer_cli_foreman/hosts/common_update_help.rb
```diff
@@ -39,13 +39,8 @@ def self.included(base)
...
           end
 
-          h.section _('Provider specific options') do |h|
-            HammerCLIForeman::Host.cr_help_extensions.each do |name, help|
-              h.section name do |h|
-                help.call(h)
-              end
-            end
-          end
+          ::HammerCLIForeman::ComputeResources.extend_help(h, :all, true )
```
mbacovsky: The named parameter needs to be named:
```ruby
::HammerCLIForeman::ComputeResources.extend_help(h, :all, add_host_specific_attrs: true )
```

mbacovsky:
@shiramax thanks a lot for the last update. It finally works as expected for my libvirt :champagne:! It is good to go, well almost. There is one thing I'd prefer not to merge now - syntax error in  common_update_help.rb which is preventing rendering host help. See comment inline.

I'm pretty happy with the rest. There are some other minor things that reserve refactoring that I would agree to merge as technical debt to unblock your kubevirt work as we discussed but also would prefer to have it fixed rather sooner then later. All are quite small things so it may be actually less work to fix them right away then creating a new PR so let me know your preference. All are marked inline but the list is:
- attribute_test.rb was not renamed
- interfaces_attrs_name_list helper method in HammerCLIForeman does not provide much value and we are duplicating data there
- I'd prefer to identify the compute resources with lowercase names to avoid issues with inconsistent naming (Ovirt vs. oVirt)

shiramax:
@mbacovsky thanks!
I've changed the syntax error and renamed the test file.
I think we can merge this PR, and I will open new PRs for your other comments.

thanks!

mbacovsky:
@shiramax thanks for the update! I'm going to merge this. The two missing things will be resolved in separate PR. Thanks for your hard work and getting it to this point. :cake:

PR #400, opened 2018-12-10
Ref #25491 - Add public key option in create compute resource CLI

shiramax:


mbacovsky:
@shiramax, thanks for the patch It looks good, could you please add a test for it?

shiramax:
thanks, @mbacovsky , please review again.

mbacovsky:
Please rebase the PR to work with latest core.

lib/hammer_cli_foreman/compute_resource.rb
```diff
@@ -70,6 +70,11 @@ def print_data(data)
 
     class CreateCommand < HammerCLIForeman::CreateCommand
 
+
+      option "--public_key_path", "PUBLIC_KEY_PATH", _("Path to a file that contains oVirt public key (For oVirt only)"), :attribute_name => :option_public_key_path,
```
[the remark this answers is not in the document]
shiramax: It seems that there are multiple parameters that are relevant to only one compute resource, I guess the reason is that we in the future we will able to add a compute resource with the same parameters.

shiramax:
@mbacovsky thanks :)
please review again.

lib/hammer_cli_foreman/compute_resource.rb
mbacovsky: If you use `:attribute_name => :option_public_key` you can remove `request_params` definition completely.

lib/hammer_cli_foreman/compute_resource.rb
```diff
@@ -82,6 +87,13 @@ class CreateCommand < HammerCLIForeman::CreateCommand
...
       failure_message _("Could not create the compute resource")
 
+      def request_params
+        params = super
+        if options["option_public_key_path"]
+          params['compute_resource']['public_key'] = options["option_public_key_path"]
+        end
+        params
+      end
       build_options
```
mbacovsky: In the command help I can see:
```
 --public-key PUBLIC_KEY                     For oVirt only
 --public_key_path PUBLIC_KEY_PATH           Path to a file that contains oVirt public key (For oVirt only)
```
This is probably unwanted. Use `build_options :without => :public_key`

lib/hammer_cli_foreman/compute_resource.rb
mbacovsky: In `--public_key_path` use dashes as in `--public-key-path`

shiramax: I tried it, but it didn't work - the public_key was not in params['compute_resource']

ofedoren:
Just few additional notes:

-  `:required => false` is not necessary and can be omitted.
-  if you specify `attribute_name => :option_public_key` then it puts the key value under `['compute_resource']` in params and `request_params` section is redundant. At least I see it in debug mode and also in the request.
-  also we need to add `--public-key-path` option to `UpdateCommand` as well to be consistent with `CreateCommand` (as mbacovsky said). And a test for it as well.
-  currently, if a path was specified with `--public-key-path` option, the specified path will be sent to the server as param instead of value in that file. While I was debugging, I noticed that when you specify `attribute_name => :option_public_key` then the `File` (or any) normalizer doesn't do its job (should we open an issue about that, @mbacovsky? It's weird, because the tests pass). I guess that's why the path is sent instead of file content.

Also I've noticed that the tests fail if there is no `request_params` section specified, which is weird also, because the actual request contains `['compute_resource']['public_key']` param. Not sure now where is the problem, in tests or core.

mbacovsky:
@shiramax, @ofedoren I did some debugging of what is going on with  `attribute_name => :option_public_key`. My findings are:
- when clamp defines an option it creates writer method for it that is used later by the parser. The writer method is called `#{attribute_name}=` and in the body the normalizer as called and the value is assigned to the attribute in the command.
- during parsing clamp finds the matching option definition and asks for the writer name and calls that method with the parsed value
- in our case --public-key-path will create writer `option_public_key=` which is later re-defined by writer for --public-key. That is the reason why the File normalizer is not triggered.
- it seems there is not a nice fix for this general behavior and we should better avoid using :attribute_name same as other existing option of different type

One solution here could be to define a new normalizer FileOrValue that will check if the value is existing file and loads its content  or use the value provided. In that case we would have to :exclude public_key and define our own with the proper normalizer and we could get rid of requiest_params modifications.
Other solution is to keep the two options as suggested in the PR but we have to fix the request params 
```ruby
params['compute_resource']['public_key'] = option_public_key_path if option_public_key_path
```
otherwise the public key is overriden with nil. 

I'm okay both ways.

What would you prefer @shiramax?

shiramax:
Thanks @mbacovsky and @ofedoren.
I fixed the request params, 
thanks!

lib/hammer_cli_foreman/compute_resource.rb
```diff
@@ -71,12 +71,21 @@ def print_data(data)
 
 
     class CreateCommand < HammerCLIForeman::CreateCommand
+      option "--public-key-path", "PUBLIC_KEY_PATH", _("Path to a file that contains oVirt public key (For oVirt only)"), :attribute_name => :option_public_key_path,
+             :required => false, :format => HammerCLI::Options::Normalizers::File.new
```
mbacovsky: I wonder why the `:required => false` and `:attribute_name => :option_public_key_path` are kept in the code. These are the default values and should have no impact on the code. Just making it more difficult to read. Is it intentional?

mbacovsky:
Excelent! Thanks @shiramax!

PR #404, opened 2018-12-23
Fixes #6159 - move image_id to be under the compute_attributes

shiramax:


mbacovsky:
[test hammer]

lib/hammer_cli_foreman/host.rb
```diff
@@ -305,6 +315,13 @@ def validate_options
         end
       end
 
+      def request_params
+        params = super
+        if option_image_id
+          HammerCLIForeman::Host.set_image_uuid(params,option_image_id)
```
mbacovsky: If we split (Single Responsibility) the `set_image_uuid` we could move getting the image to ComputeResource which would be more expected place for the code. I also don't think HammerCLIForeman::Host super-command is a good place  for storing helper methods for individual subcommands. I'd keep the assignment in the Update/Create commands. If you'd need to make it more DRY it would be better to create a module with the helper method that is included in the subcommands. 
```ruby
compute_resource_id = params['host']['compute_resource_id']
image_uuid = HammerCLIForeman::ComputeResource::Utils.image_uuid(compute_resource_id, options['option_image_id'])
params['host']['compute_attributes']['image_id'] = image uuid 
```

shiramax: we could move getting the image to ComputeResource  - which file?

mbacovsky:
@shiramax looks a way better now! I realized we have to move all the code from host.rb to the `CommonUpdateOptions` module we break the existing request_params otherwise. As a bonus we get more DRY code and `set_image_uuid` won't be misplaced.

`get_image_id` should be better named `get_image_uuid` and while it is not exactly a help util we should rename the module to just `utils` to cover both methods purpose.

shiramax:
thanks @mbacovsky, please review again

lib/hammer_cli_foreman/hosts/common_update_options.rb
```diff
@@ -78,6 +85,7 @@ def request_params
         if action == :update
           params['host']['compute_attributes']['volumes_attributes'] = nested_attributes(option_volume_list) unless option_volume_list.empty?
           params['host']['interfaces_attributes'] = interfaces_attributes unless option_interface_list.empty?
+          HammerCLIForeman::Hosts::CommonUpdateOptions::set_image_uuid(params, options["option_image_id"])  if options["option_image_id"]
```
mbacovsky: Not sure if this is the best place but we should load cr_id somewhere. How about?
```ruby
options["option_image_id"]
  compute_resource_id = params['host']['compute_resource_id'] || load_compute_resource(options["option_id"]) || <rise some error>
  params['host']['compute_attributes']['image_id'] = image_uuid(compute_resource_id, options["option_image_id"])
end
```

lib/hammer_cli_foreman/hosts/common_update_options.rb
```diff
@@ -50,9 +50,15 @@ def self.ask_password
         HammerCLI.interactive_output.ask(prompt) { |q| q.echo = false }
       end
 
+      def self.set_image_uuid(params, image_id)
```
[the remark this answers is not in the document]
shiramax: How can I load cr-id from the server? You can have multiple compute resource on the servers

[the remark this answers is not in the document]
mbacovsky: It would be just for a case of an update of an existing host. We could try to load the host and check if the CR is set there. If not we should rise usage/validation error or similar

shiramax:
thanks @mbacovsky ,
please review again.

PR #405, opened 2019-01-14
Fixes #25848 - Adding SSO func. through CLI using OpenID Connect.

rahulbajaj0509:
My attempt is to use keycloak for authenticating the user
and fetching the token, once we have the token we can send this
token to the foreman api, where we can use the `JWT` gem to
validate the token and then create a session for the user.

In this pr, I have attempted to fetch the token from keycloak using 
the rest-client gem. This token has to be passed to foreman where
it will be validated and accordingly authentication would be a success/failure.

### DIFF
--- a/config/foreman.yml
+++ b/config/foreman.yml
@@ -6,8 +6,28 @@
   :host: 'https://localhost/'
 
   # Credentials. You'll be asked for them interactively if you leave them blank here
+  # Possible values:
+
+  # Basic Auth:
+  #:default_auth_type: 'Basic_Auth'
   :username: 'admin'
   #:password: 'example'
+
+  # Oauth using the Password Grant Flow:
+  # This authentication method requires sessions to be enabled, uncomment the following
+  # lines to use this authentication method.
+  #:default_auth_type: 'Oauth_Password_Grant'
+  #:oidc_token_endpoint: https://keycloak.example.com/token
+  #:oidc_client_id: example-client-id
+
+  # Oauth using Authentication Code Flow(Two Factor):
+  # This authentication method requires sessions to be enabled, uncomment the following
+  # lines to use this authentication method.
+  #:default_auth_type: 'Oauth_Authentication_Code_Grant'
+  #:oidc_token_endpoint: https://keycloak.example.com/token
+  #:oidc_authorization_endpoint: https://keycloak.example.com/auth
+  #:oidc_client_id: example-client-id
+  #:oidc_redirect_uri: urn:ietf:wg:oauth:2.0:oob
 
   # Enable using sessions
   # When sessions are enabled, hammer ignores credentials stored in the config file
--- a/lib/hammer_cli_foreman/api/authenticator.rb
+++ b/lib/hammer_cli_foreman/api/authenticator.rb
@@ -1,0 +1,81 @@
+module HammerCLIForeman
+  module Api
+    class Authenticator
+      attr_accessor :auth_type, :uri, :settings
+      def initialize(auth_type, uri, settings)
+        @auth_type = auth_type
+        @uri = uri
+        @settings = settings
+      end
+
+      def fetch
+        if ssl_cert_authentication? && !use_basic_auth?
+          void_auth
+        elsif auth_type == AUTH_TYPES[:basic_auth]
+          basic_auth
+        elsif auth_type == AUTH_TYPES[:oauth_password_grant]
+          oauth_password_grant
+        elsif auth_type == AUTH_TYPES[:oauth_authentication_code_grant]
+          oauth_authentication_code_grant
+        end
+      end
+
+      private
+
+      def void_auth
+        VoidAuth.new(_('Using certificate authentication.'))
+      end
+
+      def basic_auth
+        if HammerCLIForeman::Sessions.enabled?
+          authenticator = InteractiveBasicAuth.new(
+            settings.get(:_params, :username) || ENV['FOREMAN_USERNAME'],
+            settings.get(:_params, :password) || ENV['FOREMAN_PASSWORD']
+          )
+          SessionAuthenticatorWrapper.new(authenticator, uri, auth_type)
+        else
+          username = settings.get(:_params, :username) || ENV['FOREMAN_USERNAME'] || settings.get(:foreman, :username)
+          password = settings.get(:_params, :password) || ENV['FOREMAN_PASSWORD']
+          if password.nil? && (username == settings.get(:foreman, :username))
+            password = settings.get(:foreman, :password)
+          end
+          InteractiveBasicAuth.new(username, password)
+        end
+      end
+
+      def oauth_password_grant
+        return unless HammerCLIForeman::Sessions.enabled?
+
+        authenticator = Oauth::PasswordGrant.new(
+          ENV['OIDC_TOKEN_ENDPOINT'] || settings.get(:foreman, :oidc_token_endpoint),
+          ENV['OIDC_CLIENT_ID'] || settings.get(:foreman, :oidc_client_id),
+          ENV['OIDC_USERNAME'],
+          ENV['OIDC_PASSWORD']
+        )
+        SessionAuthenticatorWrapper.new(authenticator, uri, auth_type)
+      end
+
+      def oauth_authentication_code_grant
+        return unless HammerCLIForeman::Sessions.enabled?
+
+        authenticator = Oauth::AuthenticationCodeGrant.new(
+          ENV['OIDC_TOKEN_ENDPOINT'] || settings.get(:foreman, :oidc_token_endpoint),
+          ENV['OIDC_AUTHORIZATION_URL'] || settings.get(:foreman, :oidc_authorization_endpoint),
+          ENV['OIDC_CLIENT_ID'] || settings.get(:foreman, :oidc_client_id),
+          ENV['OIDC_REDIRECT_URI'] || settings.get(:foreman, :oidc_redirect_uri)
+        )
+        SessionAuthenticatorWrapper.new(authenticator, uri, auth_type)
+      end
+
+      def ssl_cert_authentication?
+        (settings.get(:_params, :ssl_client_cert) || settings.get(:ssl, :ssl_client_cert)) &&
+          (settings.get(:_params, :ssl_client_key) || settings.get(:ssl, :ssl_client_key))
+      end
+
+      def use_basic_auth?
+        settings.get(:_params, :ssl_with_basic_auth) ||
+          settings.get(:ssl, :ssl_with_basic_auth)
+      end
+    end
+  end
+end
--- a/lib/hammer_cli_foreman/api/connection.rb
+++ b/lib/hammer_cli_foreman/api/connection.rb
@@ -1,14 +1,26 @@
 require 'hammer_cli_foreman/api/session_authenticator_wrapper'
+require 'hammer_cli_foreman/api/authenticator'
 require 'hammer_cli_foreman/api/interactive_basic_auth'
+require 'hammer_cli_foreman/api/oauth/authentication_code_grant'
+require 'hammer_cli_foreman/api/oauth/password_grant'
 require 'hammer_cli_foreman/api/void_auth'
+require 'uri'
 
 module HammerCLIForeman
+  CONNECTION_NAME = 'foreman'
+  AUTH_TYPES = {
+    basic_auth: 'Basic_Auth',
+    oauth_authentication_code_grant: 'Oauth_Authentication_Code_Grant',
+    oauth_password_grant: 'Oauth_Password_Grant'
+  }.freeze
+
   module Api
     class Connection < HammerCLI::Apipie::ApiConnection
       attr_reader :authenticator
 
-      def initialize(settings, logger = nil, locale = nil)
-        default_params = build_default_params(settings, logger, locale)
+      def initialize(settings, logger = nil, locale = nil, auth_type = nil)
+        auth_type ||= default_auth_type(settings)
+        default_params = build_default_params(settings, logger, locale, auth_type)
         super(default_params,
           :logger => logger,
           :reload_cache => settings.get(:_params, :reload_cache) || settings.get(:reload_cache)
@@ -16,7 +28,7 @@ module HammerCLIForeman
       end
 
       def login
-        # Call some api entry point to trigger the
+        # Call some api entry point to trigger the successful connection
         @api.resource(:home).action(:status).call
       end
 
@@ -30,40 +42,35 @@ module HammerCLIForeman
 
       protected
 
-      def create_authenticator(uri, settings)
-        return @authenticator if @authenticator
+      # If the settings in foreman.yml has use_sessions as false, use :basic_auth
+      # Else if the settings in foreman.yml has use_sessions as true
+      # and if there exists a session_file with valid contents, we use the auth_type from sessions_file
+      # Thus if the session expires (indicated by nil session_id), we use the
+      # same auth_type for re-authentication as was used by the previous session.
+      # Else we use the passed auth_type.
+      def default_auth_type(settings)
+        return AUTH_TYPES[:basic_auth] unless HammerCLIForeman::Sessions.enabled?
 
-        if ssl_cert_authentication?(settings) && !use_basic_auth?(settings)
-          @authenticator = VoidAuth.new(_('Using certificate authentication.'))
+        url = settings.get(:_params, :host) || settings.get(:foreman, :host)
+        username = settings.get(:_params, :username) || settings.get(:foreman, :username)
+        session = HammerCLIForeman::Sessions.get(url)
+        if !session.valid? && session.user_name == username && !session.auth_type.nil?
+          session.auth_type
         else
-          if settings.get(:foreman, :use_sessions)
-            @authenticator = InteractiveBasicAuth.new(
-              settings.get(:_params, :username) || ENV['FOREMAN_USERNAME'],
-              settings.get(:_params, :password) || ENV['FOREMAN_PASSWORD']
-            )
-            @authenticator = SessionAuthenticatorWrapper.new(@authenticator, uri)
-          else
-            username = settings.get(:_params, :username) || ENV['FOREMAN_USERNAME'] || settings.get(:foreman, :username)
-            password = settings.get(:_params, :password) || ENV['FOREMAN_PASSWORD']
-            if (password.nil? && (username == settings.get(:foreman, :username)))
-              password = settings.get(:foreman, :password)
-            end
-            @authenticator = InteractiveBasicAuth.new(username, password)
-          end
-          @authenticator
+          # If the caller has not sepcified an 'auth_type'
+          # and the 'default_auth_type' in settings is also undefined
+          # use :basic_auth for authentication.
+          HammerCLI::Settings.get(:foreman, :default_auth_type) || AUTH_TYPES[:basic_auth]
         end
       end
 
-      def ssl_cert_authentication?(settings)
-        (settings.get(:_params, :ssl_client_cert) || settings.get(:ssl, :ssl_client_cert)) &&
-        (settings.get(:_params, :ssl_client_key) || settings.get(:ssl, :ssl_client_key))
+      def create_authenticator(uri, settings, auth_type)
+        return @authenticator if @authenticator
+
+        @authenticator = HammerCLIForeman::Api::Authenticator.new(auth_type, uri, settings).fetch
       end
 
-      def use_basic_auth?(settings)
-        settings.get(:_params, :ssl_with_basic_auth) || settings.get(:ssl, :ssl_with_basic_auth)
-      end
-
-      def build_default_params(settings, logger, locale)
+      def build_default_params(settings, logger, locale, auth_type)
         config = {}
         config[:uri] = settings.get(:_params, :host) || settings.get(:foreman, :host)
         config[:logger] = logger unless logger.nil?
@@ -77,13 +84,11 @@ module HammerCLIForeman
         config[:timeout] = settings.get(:foreman, :request_timeout)
         config[:timeout] = -1 if (config[:timeout] && config[:timeout].to_i < 0)
         config[:apidoc_authenticated] = false
-        config[:authenticator] = create_authenticator(config[:uri], settings)
+        config[:authenticator] = create_authenticator(config[:uri], settings, auth_type)
         config
       end
     end
   end
-
-  CONNECTION_NAME = 'foreman'
 
   def self.foreman_api_connection
     HammerCLI.context[:api_connection].create(CONNECTION_NAME) do
@@ -91,8 +96,18 @@ module HammerCLIForeman
     end
   end
 
+  def self.foreman_api_reconnect(auth_type)
+    HammerCLI.context[:api_connection].drop(CONNECTION_NAME)
+    HammerCLI.context[:api_connection].create(CONNECTION_NAME) do
+      HammerCLIForeman::Api::Connection.new(
+        HammerCLI::Settings,
+        Logging.logger['API'],
+        HammerCLI::I18n.locale,
+        auth_type)
+    end
+  end
+
   def self.init_api_connection
     foreman_api_connection
   end
 end
-
--- a/lib/hammer_cli_foreman/api/oauth/password_grant.rb
+++ b/lib/hammer_cli_foreman/api/oauth/password_grant.rb
@@ -1,0 +1,81 @@
+require 'hammer_cli_foreman/openid_connect'
+
+module HammerCLIForeman
+  module Api
+    module Oauth
+      class PasswordGrant < ApipieBindings::Authenticators::TokenAuth
+        attr_accessor :oidc_token_endpoint, :oidc_client_id, :user, :password, :token
+
+        def initialize(oidc_token_endpoint, oidc_client_id, user, password)
+          @oidc_token_endpoint = oidc_token_endpoint
+          @oidc_client_id = oidc_client_id
+          @user = user
+          @password = password
+          super set_token(oidc_token_endpoint, oidc_client_id, user, password)
+        end
+
+        def authenticate(request, token)
+          if HammerCLI.interactive?
+            set_token_interactively
+          end
+          super
+        end
+
+        def set_token_interactively
+          @token ||= set_token(get_oidc_token_endpoint, get_oidc_client_id, get_user, get_password)
+        end
+
+        def set_token(input_oidc_token_endpoint, input_oidc_client_id, input_user, input_password)
+          @oidc_token_endpoint = input_oidc_token_endpoint if input_oidc_token_endpoint
+          @user = input_user
+          @password = input_password
+          @oidc_client_id = input_oidc_client_id if input_oidc_client_id
+          if @user && @password && @oidc_token_endpoint && @oidc_client_id
+            @token = HammerCLIForeman::OpenidConnect.new(
+              @oidc_token_endpoint, @oidc_client_id).get_token(@user, @password)
+          else
+            @token = nil
+          end
+        end
+
+        def error(ex)
+          if ex.is_a?(RestClient::InternalServerError)
+            @user = @password = @oidc_token_endpoint = @oidc_client_id = nil
+            original_message = _("Invalid credentials or oidc-client-id or oidc-token-endpoint.\n")
+            begin
+              message = JSON.parse(ex.response.body)['error']['message']
+            rescue
+            end
+            UnauthorizedError.new(original_message << message)
+          end
+        end
+
+        private
+
+        def get_user
+          @user ||= ask_user(_("Username:%s") % " ")
+        end
+
+        def get_password
+          @password ||= ask_user(_("Password:%{wsp}") % {:wsp => " "}, true)
+        end
+
+        def get_oidc_token_endpoint
+          @oidc_token_endpoint ||= ask_user(_("Openidc Provider Token Endpoint:%s") % " ")
+        end
+
+        def get_oidc_client_id
+          @oidc_client_id ||= ask_user(_("Client ID:%s") % " ")
+        end
+
+        def ask_user(prompt, silent=false)
+          if silent
+            HammerCLI.interactive_output.ask(prompt) { |q| q.echo = false }
+          else
+            HammerCLI.interactive_output.ask(prompt)
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/hammer_cli_foreman/api/session_authenticator_wrapper.rb
+++ b/lib/hammer_cli_foreman/api/session_authenticator_wrapper.rb
@@ -3,31 +3,29 @@ require 'uri'
 module HammerCLIForeman
   module Api
     class SessionAuthenticatorWrapper < ApipieBindings::Authenticators::Base
-      SESSION_STORAGE = '~/.hammer/sessions/'
 
-      attr_reader :session_id
+      attr_reader :session_id, :url, :auth_type
 
-      def initialize(authenticator, url, storage_dir = nil)
+      def initialize(authenticator, url, auth_type)
         @authenticator = authenticator
         @url = url
+        @auth_type = auth_type
+      end
 
-        @session_file = "#{uri.scheme}_#{uri.host}"
-        @storage_dir = storage_dir || File.expand_path(SESSION_STORAGE)
-
-        @permissions_ok = check_storage_permissions
-        warn _("Can't use session auth due to invalid permissions on session files.") unless @permissions_ok
+      def session
+        @session ||= Sessions.get(@url)
       end
 
       def clear
-        destroy_session
+        session.destroy
         @authenticator.clear if @authenticator.respond_to?(:clear)
       end
 
       def status
-        if load_session
-          _("Session exists, currently logged in as '%s'.") % @user
+        if session.valid?
+          _("Session exists, currently logged in as '%s'.") % session.user_name
         else
-          _("Using sessions, you are currently not logged in.")
+          _('Using sessions, you are currently not logged in.')
         end
       end
 
@@ -40,15 +38,13 @@ module HammerCLIForeman
       end
 
       def authenticate(request, args)
-        load_session
-
         user = @authenticator.user
 
-        @user_changed ||= (!user.nil? && user != @user)
+        @user_changed ||= (!user.nil? && user != session.user_name)
 
-        if !@user_changed && @permissions_ok && @session_id
+        if !@user_changed && Sessions.configured?(@url) && session.id
           jar = HTTP::CookieJar.new
-          jar.add(HTTP::Cookie.new('_session_id', @session_id, domain: uri.hostname.downcase, path: '/', for_domain: true))
+          jar.add(HTTP::Cookie.new('_session_id', session.id, domain: uri.hostname.downcase, path: '/', for_domain: true))
           request['Cookie'] = HTTP::Cookie.cookie_value(jar.cookies)
           request
         else
@@ -57,12 +53,11 @@ module HammerCLIForeman
       end
 
       def error(ex)
-        load_session
-        if ex.is_a?(RestClient::Unauthorized) && !@session_id.nil?
+        if ex.is_a?(RestClient::Unauthorized) && session.valid?
           if @user_changed
-            return UnauthorizedError.new(_("Invalid username or password, continuing with session for '%s'.") % @user)
+            return UnauthorizedError.new(_("Invalid username or password, continuing with session for '%s'.") % session.user_name)
           else
-            destroy_session
+            session.destroy
             return SessionExpired.new(_("Session has expired."))
           end
         else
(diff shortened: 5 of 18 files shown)


lib/hammer_cli_foreman/api/connection.rb
```diff
@@ -36,24 +37,34 @@ def create_authenticator(uri, settings)
...
-          else
-            username = settings.get(:_params, :username) || ENV['FOREMAN_USERNAME'] || settings.get(:foreman, :username)
-            password = settings.get(:_params, :password) || ENV['FOREMAN_PASSWORD']
-            if (password.nil? && (username == settings.get(:foreman, :username)))
-              password = settings.get(:foreman, :password)
-            end
-            @authenticator = InteractiveBasicAuth.new(username, password)
-          end
-          @authenticator
+          @authenticator = InteractiveTokenAuth.new(settings.get(:_params, :token) || ENV['KEYCLOAK_TOKEN'])
```
rahulbajaj0509: @mbacovsky I am trying to create an object of the class `InteractiveTokenAuth` but i am failing in doing so. Can you please take a look?

rahulbajaj0509: The whole idea is to set the @token in the `@authenticator` so that it can be passed to foreman api.

rahulbajaj0509:
@mbacovsky you need to use the command `hammer -d auth login --url "<your keycloak server url>" --realm "<your realm used in keycloak>" --username "<username>" --password "<password>"`

For username and password, you will have to create a username and password on keycloak in a defined realm. Let me know if you need any help with the setup :)

rahulbajaj0509:
@mbacovsky after https://github.com/Apipie/apipie-bindings/pull/77, I guess except the clean up required, the pr is ready for testing. It works for me, let me know if you are facing any issues while testing it.

Ah! right now it takes the expiration time of the session from foreman settings, I need to fix that!

lib/hammer_cli_foreman/api/connection.rb
```diff
@@ -1,22 +1,24 @@
...
 require 'hammer_cli_foreman/api/void_auth'
 
 module HammerCLIForeman
   module Api
     class Connection < HammerCLI::Apipie::ApiConnection
       attr_reader :authenticator
 
-      def initialize(settings, logger = nil, locale = nil)
-        default_params = build_default_params(settings, logger, locale)
+      def initialize(settings, logger = nil, locale = nil, auth_type = AuthTypes[:basic_auth])
```
mbacovsky: Can we use constants instead of the hash for auth_type? Something like AUTH_TYPE_BASIC, AUTH_TYPE_OAUTH_BASIC, AUTH_TYPE_OAUTH_2FA would be more readable.

lib/hammer_cli_foreman/commands.rb
```diff
@@ -66,7 +66,7 @@ def self.record_to_common_format(data)
   class Command < HammerCLI::Apipie::Command
 
     def self.connection_name(resource_class)
-      CONNECTION_NAME
+      HammerCLIForeman.get_connection_name
```
mbacovsky: Why do we have different  connection names for different auth methods? This name should refer to which API we are connecting and should be immutable for a resource.

rahulbajaj0509: We observed that [HammerCLIForeman.foreman_api_connection](https://github.com/theforeman/hammer-cli-foreman/pull/405/files) is being called multiple times when we run any of the login commands.

```
  def self.foreman_api_connection
    HammerCLI.context[:api_connection].create(CONNECTION_NAME) do
      HammerCLIForeman::Api::Connection.new(HammerCLI::Settings, Logging.logger['API'], HammerCLI::I18n.locale)
    end
  end
```


If we look at this method, it yields a block only if a previous connection with name {CONNECTION_NAME} (foreman) has not been called already.
From [HammerCli::Connection.rb](https://github.com/theforeman/hammer-cli/blob/c0c23e2f319f4f71c8294663a980e7830e8754e7/lib/hammer_cli/connection.rb#L21):
```
def create(name, &create_connector_block)
      unless connections[name]
        connector = yield
        @logger.debug("Registered: #{name}") if @logger
        connections[name] = connector
      end
      connections[name]
end
```

This worked fine in our initial flow since there was only a single type of authentication flow available (ie basic auth).

When we introduced more authentication ways, and called `HammerCLIForeman.foreman_api_connection` from our [auth.rb](https://github.com/theforeman/hammer-cli-foreman/blob/347bf39cf73f50830c431da99e75bdc581f0a831/lib/hammer_cli_foreman/auth.rb#L9) file, for other authentication ways, the block was never yielded, because the first time `HammerCLIForeman.foreman_api_connection` was called, the connection (with basic_auth) was already stored with `FOREMAN` key. 

Hence we did these changes, so that the block is yielded, when we use any other authentiction way except basic_auth. 

I will be glad to explain this in more detail, along with the flow as required.

lib/hammer_cli_foreman/api/connection.rb
```diff
@@ -77,22 +94,41 @@ def build_default_params(settings, logger, locale)
...
+  AuthTypes = {
+    basic_auth: "HammerCLIForeman::Auth::LoginCommand::Basic",
+    oauth_with_2fa: "HammerCLIForeman::Auth::LoginCommand::Oauth::With2fa",
+    oauth_without_2fa: "HammerCLIForeman::Auth::LoginCommand::Oauth::Without2fa"
+  }
 
-  def self.foreman_api_connection
-    HammerCLI.context[:api_connection].create(CONNECTION_NAME) do
-      HammerCLIForeman::Api::Connection.new(HammerCLI::Settings, Logging.logger['API'], HammerCLI::I18n.locale)
+  def self.foreman_api_connection(auth_type = AuthTypes[:basic_auth])
```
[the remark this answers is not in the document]
mbacovsky: The connection drop could be achieved by `HammerCLI.context[:api_connection].drop(CONNECTION_NAME)`

rahulbajaj0509:
Remaining tasks for this pr are:

---

- [ ] making the SSL CA configurable.
- [ ] convert `execute_with_params` method to a module.
- [ ] refactor the `create_authenticator` method.
- [ ] use net/http or restclient throughout.
- [ ] error handling
- [ ] update test cases for `auth.rb` and `connection.rb` 
- [ ] test cases - `keycloak.rb`, `interactive_keycloak_auth.rb`, `interactive_keycloak_auth_via_2fa.rb`. 

These tasks are the total tasks remaining for this pr, I will tick each of them once the task is complete.

lib/hammer_cli_foreman/api/connection.rb
```diff
@@ -1,22 +1,33 @@
...
+  CONNECTION_NAME = 'foreman'
+
   module Api
     class Connection < HammerCLI::Apipie::ApiConnection
       attr_reader :authenticator
 
-      def initialize(settings, logger = nil, locale = nil)
-        default_params = build_default_params(settings, logger, locale)
+      def initialize(settings, logger = nil, locale = nil, auth_type =
+          AUTH_TYPES[HammerCLI::Settings.get(:foreman, :default_auth_type).to_sym])
```
mbacovsky: if the settings does not contain the :default_auth_type I get 
`NoMethodError (undefined method `to_sym' for nil:NilClass)`. We will need some default value here.

mbacovsky:
`hammer auth status` gives me 
'Using sessions, you are currently not logged in.' all the time even when logged in.

mbacovsky:
After `hammer auth logout` I'm logged out but `auth status` gives me 'Session exists, currently logged in as ''.'

mbacovsky:
After some more login/logout excersises I'm getting correct output from `auth status` when logged in. Needs mode investigation.

lib/hammer_cli_foreman/api/connection.rb
```diff
@@ -77,22 +135,53 @@ def build_default_params(settings, logger, locale)
...
+  def self.session_repo
+    session_path = Dir.home + "/.hammer/sessions"
+    Dir[session_path].first
+  end
+
+  def self.session_file
+    if Dir[session_repo+ "/*"].empty?
+      nil
+    else
+      Dir.entries(session_repo)[-1]
```
mbacovsky: This is not safe for a couple of reasons. I have multiple session files for multiple foremans I manage in my session dir. We should pick the correct one by url, otherwise we can end up reading wrong file. 
Also while this is not predictably sorted list I ended up with '..' on the last position resulting in ugly errors when reading the session. 

I wonder if it is time to represent the session file and operations with some class and have the usage unified across the codebase.
Would something with similar interface make sense?:
```ruby
session = HammerCLIForeman::Sessions.load(url) # load or create new session
expired = session.expired?
session.expire!
session.username
session.username = username
session.auth_type
session.auth_type = auth_type
session.save
session.id
session.id =  id
```

lib/hammer_cli_foreman/api/connection.rb
rahulbajaj0509: Yeah, it is a known issue. I figured that out :) Need to fix this!

lib/hammer_cli_foreman/api/connection.rb
rahulbajaj0509: Yeah! completely makes sense. I was actually assuming that a user will have only one session file and one foreman instance to manage. Thanks for bringing this up :)

Your suggestion about making a new class that takes care of the session file and operation completely makes sense.

lib/hammer_cli_foreman/api/connection.rb
```diff
@@ -77,22 +94,41 @@ def build_default_params(settings, logger, locale)
...
-    HammerCLI.context[:api_connection].create(CONNECTION_NAME) do
-      HammerCLIForeman::Api::Connection.new(HammerCLI::Settings, Logging.logger['API'], HammerCLI::I18n.locale)
+  def self.foreman_api_connection(auth_type = AuthTypes[:basic_auth])
+    connection = get_connection_name(auth_type)
+    HammerCLI.context[:api_connection].create(connection) do
+      HammerCLIForeman::Api::Connection.new(
+        HammerCLI::Settings,
+        Logging.logger['API'],
+        HammerCLI::I18n.locale,
+        auth_type)
```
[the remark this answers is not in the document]
rahulbajaj0509: done.

ntkathole:
@rahulbajaj0509 Facing two issues:
1.
```
$ hammer auth status
Error: undefined method `status' for nil:NilClass
```

2. 
```
$ hammer -u admin -p changeme host list
Error: ApipieBindings::AuthenticatorMissingError
```

ntkathole:
@rahulbajaj0509 I face above issue when I have `:default_auth_type: basic_auth` in hammer configuration. Changing it to Basic_Auth works fine. Might be we need to handle error correctly if there is old value of deafult auth type.

rahulbajaj0509:
> @rahulbajaj0509 I face above issue when I have `:default_auth_type: basic_auth` in hammer configuration. Changing it to Basic_Auth works fine. Might be we need to handle error correctly if there is old value of deafult auth type.

ack, error handling is on my to-do list. https://github.com/theforeman/hammer-cli-foreman/pull/405#issuecomment-477321985

lib/hammer_cli_foreman/api/connection.rb
```diff
@@ -1,22 +1,37 @@
 require 'hammer_cli_foreman/api/session_authenticator_wrapper'
 require 'hammer_cli_foreman/api/interactive_basic_auth'
+require 'hammer_cli_foreman/api/oauth/interactive_keycloak_auth_via_2fa'
+require 'hammer_cli_foreman/api/oauth/interactive_keycloak_auth'
 require 'hammer_cli_foreman/api/void_auth'
+require 'uri'
 
 module HammerCLIForeman
+  AUTH_TYPES = {
```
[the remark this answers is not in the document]
rahulbajaj0509: The thought behind this hash was, we might have a number of authentication types added in the future. As the number of authentication types increases, we would have more number of constants created. Therefore, I thought of putting it to a hash and then having the hash as a constant.

Let me know your thoughts on this?

lib/hammer_cli_foreman/api/connection.rb
```diff
@@ -30,18 +45,34 @@ def login_status
...
           if settings.get(:foreman, :use_sessions)
             @authenticator = InteractiveBasicAuth.new(
               settings.get(:_params, :username) || ENV['FOREMAN_USERNAME'],
               settings.get(:_params, :password) || ENV['FOREMAN_PASSWORD']
             )
-            @authenticator = SessionAuthenticatorWrapper.new(@authenticator, uri)
+            @authenticator = SessionAuthenticatorWrapper.new(@authenticator, uri, auth_type)
           else
             username = settings.get(:_params, :username) || ENV['FOREMAN_USERNAME'] || settings.get(:foreman, :username)
             password = settings.get(:_params, :password) || ENV['FOREMAN_PASSWORD']
```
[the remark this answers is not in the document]
rahulbajaj0509: From what I have understood is that a user might be logged in through `settings.get(:_params, :username) || ENV['FOREMAN_USERNAME']`. Now in this case, if the password is not provided then it should be asked interactively to the user and not taken from `settings.get(:foreman, :password)`.

There will always be a user and password by default in the settings file, so we want to make sure that if we are using a user from the settings file only then we use the password from the settings file. 

If we do something like adding the condition `(settings.get(:foreman, :password)` to the above line then it might happen that a user from `settings.get(:_params, :username) || ENV['FOREMAN_USERNAME']` catch the password from the settings which would be wrong.

Let me know your thoughts on this.

lib/hammer_cli_foreman/api/connection.rb
```diff
@@ -50,7 +81,41 @@ def create_authenticator(uri, settings)
...
+              settings.get(:_params, :password) || ENV['KEYCLOAK_PASSWORD']
+            )
+            @authenticator = SessionAuthenticatorWrapper.new(@authenticator, uri, auth_type)
+          else
+            url = settings.get(:_params, :url) || ENV['KEYCLOAK_URL'] || settings.get(:foreman, :url)
+            realm = settings.get(:_params, :realm) || ENV['KEYCLOAK_REALM'] || settings.get(:foreman, :realm)
+            client_id = settings.get(:_params, :client_id) || ENV['KEYCLOAK_CLIENT_ID'] || settings.get(:foreman, :client_id)
+            username = settings.get(:_params, :username) || ENV['KEYCLOAK_USERNAME'] || settings.get(:foreman, :username)
+            password = settings.get(:_params, :password) || ENV['KEYCLOAK_PASSWORD'] || settings.get(:foreman, :password)
+            if (password.nil? && (username == settings.get(:foreman, :username)))
```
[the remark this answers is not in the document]
rahulbajaj0509: I think the above condition in line 99 should be removed, good catch :+1:

lib/hammer_cli_foreman/api/connection.rb
```diff
@@ -30,18 +45,34 @@ def login_status
...
+        use_session = settings.get(:foreman, :use_sessions)
+        if use_session && !HammerCLIForeman.session_contents.empty? &&
+          HammerCLIForeman.session_contents[:session_id].nil?
+          HammerCLIForeman.session_contents[:auth_type]
+        else
+          auth_type
+        end
+      end
+
+      def create_authenticator(uri, settings, auth_type)
```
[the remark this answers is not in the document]
rahulbajaj0509: Done :) Need to fix the tests though, will do that shortly.

lib/hammer_cli_foreman/api/session_authenticator_wrapper.rb
```diff
@@ -72,81 +67,35 @@ def error(ex)
 
       def response(r)
         if (r.cookies['_session_id'] && r.code != 401)
-          @session_id = r.cookies['_session_id']
-          save_session(@session_id, @authenticator.user)
+          session.id = r.cookies['_session_id']
+          session.user_name = @authenticator.user
```
mbacovsky: Here I believe we need to store also the auth type:
```ruby
session.auth_type = @auth_type
```

lib/hammer_cli_foreman/sessions.rb
```diff
@@ -0,0 +1,101 @@
...
+        true
+      end
+    end
+  end
+
+  class Session
+    attr_accessor :id, :user_name, :auth_type
+
+    def initialize(session_path)
+      @session_path = session_path
```
mbacovsky: The session_path should be expanded otherwise the session is stored in CWD
```ruby
session_path = File.expand_path(session_path)
```

lib/hammer_cli_foreman/api/session_authenticator_wrapper.rb
```diff
@@ -3,29 +3,27 @@
 module HammerCLIForeman
   module Api
     class SessionAuthenticatorWrapper < ApipieBindings::Authenticators::Base
-      SESSION_STORAGE = '~/.hammer/sessions/'
```
mbacovsky: SESSION_STORAGE = File.expand_path('~/.hammer/sessions/')

lib/hammer_cli_foreman/sessions.rb
```diff
@@ -0,0 +1,101 @@
+require 'uri'
+module HammerCLIForeman
+  class Sessions
+
+    STORAGE_DIR = '~/.hammer/sessions/'
```
mbacovsky: SESSION_STORAGE = File.expand_path('~/.hammer/sessions/')

lib/hammer_cli_foreman/sessions.rb
mbacovsky: It is better to get corrected above so that STORAGE_DIR is expanded since beginning

lib/hammer_cli_foreman/api/authenticator.rb
```diff
@@ -0,0 +1,96 @@
...
+    class Authenticator
+      attr_accessor :auth_type, :uri, :settings
+      def initialize(auth_type, uri, settings)
+        @auth_type = auth_type
+        @uri = uri
+        @settings = settings
+      end
+
+      def fetch
+        if ssl_cert_authentication? && !use_basic_auth
```
mbacovsky: `use_basic_auth` does not exist in the context. It probably was `use_basic_auth?` and needs to be moved to Authenticator

lib/hammer_cli_foreman/api/session_authenticator_wrapper.rb
```diff
@@ -40,15 +38,13 @@ def user_changed?
...
       def authenticate(request, args)
-        load_session
-
         user = @authenticator.user
 
-        @user_changed ||= (!user.nil? && user != @user)
+        @user_changed ||= (!user.nil? && user != session.user_name)
 
-        if !@user_changed && @permissions_ok && @session_id
+        if !@user_changed && @permissions_ok && session.id
```
mbacovsky: @rahulbajaj0509 here we should decide if @permissions_ok should be replaced with something like `Sessions.configured?` and block using sessions or if we are okay with the warning. One way or another @permissions_ok was removed and is always unsed resulting in not using sessions at all.

lib/hammer_cli_foreman/api/session_authenticator_wrapper.rb
```diff
@@ -72,81 +67,35 @@ def error(ex)
...
         @authenticator.password(ask) if @authenticator.respond_to?(:password)
       end
 
-      def set_credentials(*args)
-        @authenticator.set_credentials(*args) if @authenticator.respond_to?(:set_credentials)
-      end
-
-      protected
-
-      def uri
```
mbacovsky: this is still used in creating the CookieJar

lib/hammer_cli_foreman/api/session_authenticator_wrapper.rb
```diff
@@ -57,12 +53,11 @@ def authenticate(request, args)
       end
 
       def error(ex)
-        load_session
-        if ex.is_a?(RestClient::Unauthorized) && !@session_id.nil?
+        if ex.is_a?(RestClient::Unauthorized) && !session.valid?
```
mbacovsky: `&& session.valid?` the session needs to be valid before we destroy it. When the session is not valid in time we get Unauthorized error we probably provided wrong credentials or there is some other reason for 401.

mbacovsky:
@rahulbajaj0509, when I added all the changes mentioned above the most used flows seems to be working. It need very careful testing though as I found errors also in other auth methods and we need to make sure we don't  introduce regressions in auth.

rahulbajaj0509:
@mbacovsky thanks for comments, while resolving them I realized we have no condition such that deletes the session file once created.

mbacovsky:
@rahulbajaj0509, hmm, I'm not sure what condition that would be we are reusing the same file over and over. Also I'm not sure if we need one. The number of session files equals to number of servers you have connected to and the number will be very small. Any concerns I miss?

rahulbajaj0509:
> @rahulbajaj0509, hmm, I'm not sure what condition that would be we are reusing the same file over and over. Also I'm not sure if we need one. The number of session files equals to number of servers you have connected to and the number will be very small. Any concerns I miss?

I remember we had this discussion and it makes sense to not delete the file, I just wanted to make sure that we didn't miss anything :)

ofedoren:
@rahulbajaj0509, let me add a few additional concerns (please mark if resolved):

  I. With sessions disabled:
  - [ ] `hammer host list` or similar command causes `ApipieBindings::AuthenticatorMissingError`.
  - [ ] `hammer auth status` causes `undefined method 'status' for nil:NilClass`.
  
  II. Just refactoring/UX stuff:
  - [ ] `hammer auth login oauth -h` shows inconsistent options like `-client-id` or `-two-factor`. We use one dash `-` for single character options like `-u` stands for `--user` etc. You can simply keep full versions like `--two-factor` or in case of `--client-id` add `-c` which stands for `--client-id`.
  - [ ] `default_auth_type` entry in `foreman.yml` config file should contain all possible (supported) values in comment like https://github.com/theforeman/hammer-cli/blob/master/config/cli_config.template.yml#L29-L33.

lib/hammer_cli_foreman/api/authenticator.rb
```diff
@@ -0,0 +1,96 @@
...
+        if HammerCLIForeman::Sessions.enabled?
+          authenticator = Oauth::InteractiveKeycloakAuth.new(
+            settings.get(:_params, :url) || ENV['KEYCLOAK_URL'],
+            settings.get(:_params, :realm) || ENV['KEYCLOAK_REALM'],
+            settings.get(:_params, :client_id) || ENV['KEYCLOAK_CLIENT_ID'],
+            settings.get(:_params, :username) || ENV['KEYCLOAK_USERNAME'],
+            settings.get(:_params, :password) || ENV['KEYCLOAK_PASSWORD']
+          )
+          SessionAuthenticatorWrapper.new(authenticator, uri, auth_type)
+        else
```
ofedoren: This part of the condition will not work (dead code) for oauth with/out because of https://github.com/theforeman/hammer-cli-foreman/blob/f0c3cecb34dc897dc4931a3b29f3a5cb7a118f74/lib/hammer_cli_foreman/api/connection.rb#L53

It's always `basic_auth` if sessions are disabled. Please, correct me if I'm wrong.

lib/hammer_cli_foreman/api/connection.rb
```diff
@@ -1,22 +1,35 @@
...
+require 'hammer_cli_foreman/api/authenticator'
 require 'hammer_cli_foreman/api/interactive_basic_auth'
+require 'hammer_cli_foreman/api/oauth/interactive_keycloak_auth_via_2fa'
+require 'hammer_cli_foreman/api/oauth/interactive_keycloak_auth'
 require 'hammer_cli_foreman/api/void_auth'
+require 'uri'
 
 module HammerCLIForeman
+  AUTH_TYPES = {
+    basic_auth: "Basic_Auth",
```
ofedoren: We should have documented those in `foreman.yml` so users know what exactly they should specify.

lib/hammer_cli_foreman/api/authenticator.rb
```diff
@@ -0,0 +1,96 @@
...
+
+      def ouath_with_2fa
+        if HammerCLIForeman::Sessions.enabled?
+          authenticator = Oauth::InteractiveKeycloakAuthVia2fa.new(
+            settings.get(:_params, :url) || ENV['KEYCLOAK_URL'],
+            settings.get(:_params, :realm) || ENV['KEYCLOAK_REALM'],
+            settings.get(:_params, :client_id) || ENV['KEYCLOAK_CLIENT_ID']
+          )
+          SessionAuthenticatorWrapper.new(authenticator, uri, auth_type)
+        else
```
ofedoren: Same dead code.

Also, `settings.get(:_params, :url/realm/etc)` will not work and don't make sense to have it because `_params` will contain options for `hammer` (main) command parsed from user's input in console/terminal. Although `settings.get(:_params, :username/password)` will work, see https://github.com/theforeman/hammer-cli/blob/8ba92b7375dc31934698d9db0d1f3399037a0011/bin/hammer#L80.

ofedoren:
Functional-wise it seems to be working with sessions enabled/disabled for basic/oauth authentications for me. Didn't check certs only authentication though...

lib/hammer_cli_foreman/api/authenticator.rb
rahulbajaj0509: Correct, it's always `basic_auth` if sessions are disabled. The pointed code above is dead code.

rahulbajaj0509:
@ofedoren @mbacovsky @ntkathole 

I have made all the required changes. I have tested the PR and everything works fine. I will squash the commits once I have your approval :smile:

mbacovsky:
@rahulbajaj0509, could you please rebase the PR?  It seems to be based on hammer 0.16 and I'd like to check it works well with the recent stuff.

lib/hammer_cli_foreman/api/oauth/interactive_keycloak_auth.rb
```diff
@@ -0,0 +1,89 @@
...
+          if @user && @password && @realm && @openidc_provider_url && @client_id
+            @token = HammerCLIForeman::Keycloak.new(
+              input_openidc_provider_url, input_realm, input_client_id).get_token(input_user, input_password)
+          else
+            @token = nil
+          end
+        end
+
+        def error(ex)
+          # if ex.is_a?(RestClient::Unauthorized)
```
mbacovsky: Why this is commented out?

lib/hammer_cli_foreman/auth.rb
```diff
@@ -1,34 +1,69 @@
-module HammerCLIForeman
+require 'hammer_cli_foreman/keycloak.rb'
```
mbacovsky: When we decided to make the integration for generic OAuth shouldn't we get rid of Keycloak being mentioned in the code? Do we have any Keycloak specific code in here or we can rename it to OAuth across the codebase?

lib/hammer_cli_foreman/auth.rb
```diff
@@ -1,34 +1,69 @@
...
+        extend HammerCLIForeman::Authenticate::Login
 
-        print_message(_("Successfully logged in as '%s'.") % HammerCLIForeman.foreman_api_connection.authenticator.user)
-        HammerCLI::EX_OK
+        command_name('oauth')
+        desc('supports for both with/without 2fa')
+
+        option ["-u", "--username"], "USERNAME", _("Username to access the remote system")
+        option ["-p", "--password"], "PASSWORD", _("Password to access the remote system")
+        option ["-o", "--openidc-provider-url"], "URL", _("Url for keycloak server")
```
mbacovsky: Shouldn't we follow the same model in the description - "Url of the OpenID Connect provider"? and rid of mentioning Keycloak?

lib/hammer_cli_foreman/keycloak.rb
```diff
@@ -0,0 +1,52 @@
...
+  class Keycloak
+
+    def initialize(url, realm, client_id)
+      @url = url
+      @realm = realm
+      @client_id = client_id
+    end
+
+    def get_token(username, password)
+      uri = URI.parse(token_url)
```
mbacovsky: Could we add some logging in here saying what is the url we use to get token and what contained the response.

lib/hammer_cli_foreman/api/connection.rb
```diff
@@ -1,22 +1,35 @@
...
         super(default_params,
           :logger => logger,
           :reload_cache => settings.get(:_params, :reload_cache) || settings.get(:reload_cache)
         )
       end
 
       def login
-        # Call some api entry point to trigger the
+        # Call some api entry point to trigger the successful connection
         @api.resource(:home).action(:status).call
```
mbacovsky: It would be worth considering to choose different call. The status became quite expensive recently.  It touches a lot of resources especially when Katello is around. :auth_statuses :index looked promising or :architectures.

lib/hammer_cli_foreman/auth.rb
```diff
@@ -1,34 +1,70 @@
...
+        extend HammerCLIForeman::Authenticate::Login
 
-        print_message(_("Successfully logged in as '%s'.") % HammerCLIForeman.foreman_api_connection.authenticator.user)
-        HammerCLI::EX_OK
+        command_name('oauth')
+        desc('supports for both with/without 2fa')
+
+        option ["-u", "--username"], "USERNAME", _("Username to access the remote system")
+        option ["-p", "--password"], "PASSWORD", _("Password to access the remote system")
+        option ["-ote", "--openidc-token-endpoint"], "URL", _("Openidc Provider URL which issues access token")
```
ofedoren: This will not work, since clamp uses `-` for single character options.

I would suggest:
 - `-t` for `--openidc-token-endpoint`
 - `-a` for `--openidc-authorizarion-endpoint`
 - `-f` for `--two-factor`
 - `-r` for `--redirect-uri`

lib/hammer_cli_foreman/api/connection.rb
mbacovsky: @ofedoren you are right I didn't noticed there are two endpoints and was mislead by other error.

lib/hammer_cli_foreman/api/authenticator.rb
```diff
@@ -0,0 +1,81 @@
...
+          end
+          InteractiveBasicAuth.new(username, password)
+        end
+      end
+
+      def oauth_password_grant
+        return unless HammerCLIForeman::Sessions.enabled?
+
+        authenticator = Oauth::PasswordGrant.new(
+          ENV['OIDC_OPENIDC_TOKEN_ENDPOINT'],
```
mbacovsky: Isn't the _OPENIDC redundant with the OIDC prefix?

mbacovsky: Anyway to make the settings work here we need 
```ruby
ENV['OIDC_OPENIDC_TOKEN_ENDPOINT'] || settings.get(:foreman, :openidc_token_endpoint),
```

lib/hammer_cli_foreman/api/authenticator.rb
```diff
@@ -0,0 +1,81 @@
...
+        SessionAuthenticatorWrapper.new(authenticator, uri, auth_type)
+      end
+
+      def oauth_authentication_code_grant
+        return unless HammerCLIForeman::Sessions.enabled?
+
+        authenticator = Oauth::AuthenticationCodeGrant.new(
+          ENV['OPENIDC_TOKEN_ENDPOINT'] || settings.get(:foreman, :openidc_token_endpoint),
+          ENV['OPENIDC_AUTHORIZATION_URL'] || settings.get(:foreman, :openidc_authorization_endpoint),
+          ENV['OPENIDC_CLIENT_ID'] || settings.get(:foreman, :client_id),
```
mbacovsky: why client_id and redirect_id don't have the `openidc_` prefix?

lib/hammer_cli_foreman/api/oauth/password_grant.rb
```diff
@@ -0,0 +1,82 @@
...
+          if @user && @password && @openidc_token_endpoint && @client_id
+            @token = HammerCLIForeman::OpenidConnect.new(
+              @openidc_token_endpoint, @client_id).get_token(@user, @password)
+          else
+            @token = nil
+          end
+        end
+
+        def error(ex)
+          # if ex.is_a?(RestClient::Unauthorized)
```
mbacovsky: This method is handler of attempt to login with PasswordGrant. If you provide wrong credentials or client id or token endpoint it endsup here with RestClient::InternalServerError. We need to catch it and and change to UnauthorizedError with message useful to the user in his context.
We were not able to get the token and sent empty server returned "Nil JSON web token" and it is not useful unless you know what is going on under the hood. Similar it should be for code grant.

mbacovsky:
@rahulbajaj0509 the failing tests are related. It fails on missing  TokenAuth from apipie-bindings. I'll have to do a new reelase of bindings and you will have to update dependencies of h-c-f. That should fix it.

config/foreman.yml
```diff
@@ -6,9 +6,29 @@
   :host: 'https://localhost/'
 
   # Credentials. You'll be asked for them interactively if you leave them blank here
+  # Possible values:
+
+  # Basic Auth:
+  #default_auth_type: 'Basic_Auth'
```
ofedoren: This should be `#:default_auth_type:` also below for PGF and ACF. Notice the coulmn before `default...`.

If user will just uncomment those lines, it will not work without the column (parser will not treat it like a symbol IIUC).

rahulbajaj0509: good catch!

ofedoren:
Just one more thing. Is it okay that if user uses two factor auth and if wants to check status via `hammer auth status` it will ask him the code again?..

lib/hammer_cli_foreman/api/oauth/password_grant.rb
```diff
@@ -0,0 +1,81 @@
...
+            @token = HammerCLIForeman::OpenidConnect.new(
+              @oidc_token_endpoint, @oidc_client_id).get_token(@user, @password)
+          else
+            @token = nil
+          end
+        end
+
+        def error(ex)
+          if ex.is_a?(RestClient::InternalServerError)
+            self.clear
```
mbacovsky: When logging in with invalid password I get 
[ERROR 2019-10-25T15:37:22 Exception] Error: undefined method `clear' for #<HammerCLIForeman::Api::Oauth::PasswordGrant:0x0000000002fb0098>
Error: undefined method `clear' for #<HammerCLIForeman::Api::Oauth::PasswordGrant:0x0000000002fb0098>

mbacovsky: The clear should probably remove the values from the object

ofedoren: And that's why we need two reviewers... When I looked at the code I didn't notice that there is no method `clear` for `AuthenticationCodeGrant` and `PasswordGrant` objects.

@rahulbajaj0509, is it a leftover or you forget to define those methods for `PasswordGrant` and `AuthenticationCodeGrant` objects? This method exists only in `InteractiveBasicAuth`.

mbacovsky:
@rahulbajaj0509 thanks for all the amazing effort you've put into this feature!
@ofedoren thanks for testing and helpful reviews!

rahulbajaj0509:
Thanks a lot for your support @mbacovsky. I was stuck so many times when thinking about foreman_ap_reconnect, sessions, getting the correct flows in order and it was all possible because of your constant push. 

I made so many mistakes and thanks for catching them all @ofedoren! :P thanks for testing the PR again and again. I know it can be really tedious to test a PR of this size repeatedly.

Thanks a lot @mbacovsky @ofedoren for your valuable inputs, without you folks I wouldn't have made it :smile:

ntkathole:
@rahulbajaj0509 @mbacovsky 
I have sessions true in hammer config and
```

$ hammer host list
Warning: An error occured while loading module hammer_cli_foreman_remote_execution.
[Foreman] Username: admin
[Foreman] Password for admin: 
undefined local variable or method `logger' for #<Foreman::BruteforceProtection:0x000000000c89ba28>
```

Traceback - https://paste.fedoraproject.org/paste/NDIeSMgm-uYi9QkOT0rCsw

I am not sure this is related to PR but, how can I resolve it?

lib/hammer_cli_foreman/auth.rb
```diff
@@ -1,32 +1,67 @@
...
+        command_name('oauth')
+        desc('supports for both with/without 2fa')
+
+        option ["-u", "--username"], "USERNAME", _("Username to access the remote system")
+        option ["-p", "--password"], "PASSWORD", _("Password to access the remote system")
+        option ["-t", "--oidc-token-endpoint"], "OPENIDC-TOKEN-ENDPOINT", _("Openidc provider URL which issues access token")
+        option ["-a", "--oidc-authorization-endpoint"], "OPENIDC-AUTHORIZATION-ENDPOINT", _("Openidc provider URL which issues authentication code")
+        option ["-c", "--oidc-client-id"], "OPENIDC-CLIENT-ID", _("Client id used in the Openidc provider")
+        option ["-f", "--two-factor"], :flag, _("Authenticate with two factor")
+        option ["-r", "--oidc-redirect-uri"], "OPENIDC-REDIRECT-URI", _("Redirect URI for the authencation code grant flow")
```
ntkathole: typo s/authencation/authentication

rahulbajaj0509: good catch!

rahulbajaj0509:
@ntkathole thanks for having a look :+1: 

> @rahulbajaj0509 @mbacovsky
> 
>     1. I have sessions true in hammer config and
> 
> 
> ```
> 
> $ hammer host list
> Warning: An error occured while loading module hammer_cli_foreman_remote_execution.
> [Foreman] Username: admin
> [Foreman] Password for admin: 
> undefined local variable or method `logger' for #<Foreman::BruteforceProtection:0x000000000c89ba28>
> ```
> 

This issue is not reproducible in my Foreman instance. Let me help you with this once I am back :)

> Traceback - https://paste.fedoraproject.org/paste/NDIeSMgm-uYi9QkOT0rCsw
> 
> I am not sure this is related to PR but, how can I resolve it?
> 
> 
> ```
> $ hammer auth login oauth
> Warning: An error occured while loading module hammer_cli_foreman_remote_execution.
> Openidc Provider Token Endpoint: 
> Client ID: 
> Username: 
> Password: 
> Error: undefined method `request_uri' for #<URI::Generic >
> ```

Good catch! I will fix this :)

PR #410, opened 2019-02-20
Fixes #26099 - Add bootdisk back to params

chris1984:
The pr that changed this has caused regression, I have used this for my fog-vsphere testing and @lzap is hitting this with a customer too. Opening a pr to revert the change. I have not seen any regression with this when doing testing etc.

### DIFF
--- a/lib/hammer_cli_foreman/hosts/common_update_options.rb
+++ b/lib/hammer_cli_foreman/hosts/common_update_options.rb
@@ -36,10 +36,12 @@ module HammerCLIForeman
           :format => HammerCLI::Options::Normalizers::KeyValueList.new
         base.option "--interface", "INTERFACE", _("Interface parameters"), :multivalued => true,
           :format => HammerCLI::Options::Normalizers::KeyValueList.new
+        base.option "--provision-method", "METHOD", " ",
+          :format => HammerCLI::Options::Normalizers::Enum.new(['build', 'image', 'bootdisk'])
 
         base.build_options :without => [
               # - temporarily disabled params that will be removed from the api ------------------
-              :capabilities, :flavour_ref, :image_ref, :start,
+              :provision_method, :capabilities, :flavour_ref, :image_ref, :start,
               :network, :cpus, :memory, :provider, :type, :tenant_id, :image_id,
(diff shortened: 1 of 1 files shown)


chris1984:
@mbacovsky can I get a review please?

mbacovsky:
@chris1984 Why it was fixed here and not in the API? The fix should be done there IMO (regardless of the fact that this should be added by the bootdisk plugin).

After some investigation I found out with recent Foreman versions it correctly lists also the `bootdisk` option. https://github.com/theforeman/foreman/commit/51685ba80898aa6835a2396c14c02d8af52aec13 this adds the capability to extend provision methods from plugins. It is there for quite some time (since Foreman 1.11). So the question is why the option is not present in Sat. I was even not able to get it working in latest Sat builds but upstream works just fine for me (tried all with same hammer binary). All and all I'd prefer to find the cause then introducing a workaround here.

chris1984:
@mbacovsky hey sounds good I will close this PR here and work with @lzap to see why we are hitting the issue downstream since I am not that familar with the bootdisk plugin other than using it as a user.

lzap:
Hey, it works upstream for me, however I was not able test this end to end as I don't have my VMWare setup. All you need to do is install/enable bootdisk plugin and the bootdisk option appears:

```
 --provision-method PROVISION_METHOD       The method used to provision the host.
                                           Possible value(s): 'build', 'image', 'bootdisk'
```

I am able to submit new host but it will fail since I don't have a VMWare CR:

```
[lzap@box hammer-cli]$ hammer-quick host create --name host$RANDOM --organization MyOrg --location MyLoc --hostgroup "CentOS 7.1" --interface mac=AA:BB:CC:DD:EE:A2  --provision-method bootdisk
Could not create the host:
  Ip can't be blank
  Ip6 can't be blank
  Provision method is an unsupported provisioning method, available: build
```

lzap:
I've been able to find what fixed this in core: https://github.com/theforeman/foreman/pull/6399

Early class loading caused APIPIE to generate API chache before bootdisk plugin got registered. Ivan's patch fixed it. Therefore this is fixed in 1.21.

PR #412, opened 2019-03-05
Fixes #26234 - move compute resource validation to their class

shiramax:
This PR includes: 
- Changing the validation in case of creating a new compute resource to their class, which makes the registration of each compute resource very easy.
-  Changing the compute resource name to lower case letter
- Remove interfaces_attrs_name_list helper method in HammerCLIForeman

### DIFF
--- a/lib/hammer_cli_foreman/compute_resource.rb
+++ b/lib/hammer_cli_foreman/compute_resource.rb
@@ -1,4 +1,6 @@
 require 'hammer_cli_foreman/image'
+require 'hammer_cli_foreman/compute_resource/register_compute_resources'
+
 
 module HammerCLIForeman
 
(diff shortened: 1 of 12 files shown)


lib/hammer_cli_foreman/compute_resource.rb
```diff
@@ -70,24 +72,15 @@ def print_data(data)
...
       failure_message _("Could not create the compute resource")
 
       build_options
 
       validate_options do
-        required_options = (VALIDATION_PER_PROVIDER[
+        required_options = (::HammerCLIForeman.compute_resources[
           option(:option_provider).value.to_s.downcase
-        ] || []).unshift(:option_name, :option_provider).uniq
+        ] || []).compute_resource_creation_validation
```
mbacovsky: We've lost `:option_name` and `:option_provider` was it intentional? If not can we add it to the base compute_resource and extend in successors?

mbacovsky: can we make `compute_resource_creation_validation` more generic? E.g. 
```ruby 
def mandatory_attributes; [:name, :provider]; end
```
and use it like 
```ruby
... .mandatory_attributes.map { |attr| "option_#{attr}".to_sym }
```

mbacovsky: I see now the name I suggested may lead to confusion with host attributes and understand why you chose the name. Maybe `mandatory_resource_options` could help to avoid the `validation` in the name which is suggesting it is validating something?

shiramax: @mbacovsky
 can you explain how the mandatory_resource_options could help avoid the validation in the name?
thanks.

mbacovsky: 'mandatory_resource_options' would be new name for `compute_resource_creation_validation`. 
In the base class you'd have
```ruby
def mandatory_resource_options; [:name, :provider]; end
```
in the libvirt it would be
```ruby
def mandatory_resource_options; super + [:url]; end
```
etc. And in the command it would be
```ruby
      validate_options do
         required_options = (::HammerCLIForeman.compute_resources[
￼             option(:option_provider).value.to_s.downcase		￼           
￼           ] || []).mandatory_resource_options.map { |attr| "option_#{attr}".to_sym }
         all(*required_options).required
￼       end
```
Makes sense now?

lib/hammer_cli_foreman/compute_resource/ec2.rb
```diff
@@ -14,7 +14,11 @@ def compute_attributes
             'managed_ip'
         ]
       end
+
+      def mandatory_resource_options
+        super + [:name, :provider, :region, :user, :password]
```
mbacovsky: Extra `:name` and `:provider`?

mbacovsky:
@shiramax thanks for the update. I've found that in the EC2 resource the mandatory options are doubled, see inline. Except that it is good to go!

mbacovsky:
@shiramax thanks for the update. All my comments were addressed.

PR #416, opened 2019-04-04
Fixes #26536 - Possibility to change host loc/org via hammer

ofedoren:


ofedoren:
@akofink, try `hammer host update --name myhost.example.com --organization-id id_of_hosts_current_organization --new-organization-id id_of_organization_to_be_assigned`

akofink:
Great! This scenario isn't currently covered by tests afaict, right?

ofedoren:
@akofink, I'm afraid it isn't (or there are some minor oversights like in https://github.com/Katello/hammer-cli-katello/pull/645). Also I've just noticed, when tried this command, that `name` passed as a parameter under `host` scope which isn't quite good since name should remain the same (command with those parameters should not chande the host's name). It will not change the host's name since it is the same, but still not quite good...

mbacovsky:
[test hammer]

akofink:
What's going on here?

swadeley:
@ofedoren why no Assignees?

ofedoren:
@akofink, @mbacovsky currently isn't available :(

@swadeley, it's not so important in the scope of github. There is an assignee in the redmine issue.

lib/hammer_cli_foreman/hosts/common_update_options.rb
```diff
@@ -77,6 +77,20 @@ def request_params
         if action == :update
           params['host']['compute_attributes']['volumes_attributes'] = nested_attributes(option_volume_list) unless option_volume_list.empty?
           params['host']['interfaces_attributes'] = interfaces_attributes unless option_interface_list.empty?
+          new_location = option_new_location_name || option_new_location_title
```
mbacovsky: Could this logic be extracted to the OptionSource so that it is resolved along with other options and reusable like the new OptionBuilder

ofedoren:
@mbacovsky, updated.

Now there are new [option sources](https://github.com/theforeman/hammer-cli-foreman/blob/63653cd2c1ff12e2e2fd3cf75df3cfb5cbade5d0/lib/hammer_cli_foreman/option_sources/new_params.rb) that will resolve any `new_resource_name` option.

mbacovsky:
@ofedoren, I like how the new option source is designed. Would you mind fixing the conflict before I do last round of testing?

ofedoren:
@mbacovsky, done :)

mbacovsky:
Thanks @ofedoren!  :champagne:

PR #417, opened 2019-04-08
Fixes #24717 - Create hostgroup with puppet classes

ofedoren:


mbacovsky:
@ofedoren, looks good! Would it be possible and reasonable to add a test which mocks situation when there is more then a 1000 records  searched by name? The goal would be to test the API is called multiple times and the results are merged.

ofedoren:
@mbacovsky, updated. I've slightly changed the `retrieve_all` method, but the logic still remains the same. A test for it added as well.

ofedoren:
This also solves the https://bugzilla.redhat.com/show_bug.cgi?id=1550487 issue.

PR #409, opened 2019-02-11
Fixes #26008 - parameter_type option to parameter commands

kgaikwad:

PR #418, opened 2019-04-09
Refs #26008 - parameter type to parameter commands

ezr-ondrej:
Adds parameter_type.

I have realized it is going to be addressed already in the middle of the process.
I have found @kgaikwad work, but as in the process, I realized we need to do a bit more anyway.

Sorry @kgaikwad for being sloppy and not looking before starting to code :(
But I would suggest merging #409 first and I will rebase my work on that one, as there is default value addressed, that I forgot and should be in other resources as well.


ofedoren:
@ezr-ondrej, https://github.com/theforeman/hammer-cli-foreman/pull/409 is merged, so this needs a rebase.

kgaikwad:
@ezr-ondrej,
no worries! :-) 
:+1:  for extraction of data types into a constant.

ezr-ondrej:
@ofedoren yeah, I believe it's ready :)

ofedoren:
@ezr-ondrej, merging then, thanks! :)

PR #419, opened 2019-04-10
[CP] Picks for 0.15.2

jturel:
This is in support of Katello 3.10.1 => https://github.com/Katello/hammer-cli-katello/pull/639

### DIFF
--- a/hammer_cli_foreman.gemspec
+++ b/hammer_cli_foreman.gemspec
@@ -25,7 +25,7 @@ EOF
   s.extra_rdoc_files = Dir['{doc}/**/*', 'README*']
   s.require_paths = ["lib"]
 
-  s.add_dependency 'hammer_cli', '>= 0.15.1'
+  s.add_dependency 'hammer_cli', '>= 0.15.1', '< 0.16.0'
   s.add_dependency 'apipie-bindings', '>= 0.2.2'
   s.add_dependency 'rest-client', '>= 1.8.0', '< 3.0.0'
 
--- a/lib/hammer_cli_foreman/version.rb
+++ b/lib/hammer_cli_foreman/version.rb
@@ -1,5 +1,5 @@
 module HammerCLIForeman
   def self.version
-    @version ||= Gem::Version.new '0.15.1'
+    @version ||= Gem::Version.new '0.15.2'
   end
 end


jturel:
[test hammer]

jturel:
@mbacovsky can you help me out with a release for this?

jturel:
@tbrisker may I interest you in doing a release for hammer-cli-foreman?

mbacovsky:
Just for a record: we agreed on limiting deps on hammer-cli < 0.16 and releasing a new gem. It is not necessary to do a new rpm/deb release as the issue can be only reproduced on gem based installs.

jturel:
If I'm understanding the test failure right, I think the culprit is:

https://github.com/theforeman/foreman-infra/blob/50d36964d0e2b286e9466e7ebce0feae7e6b014e/puppet/modules/jenkins_job_builder/files/theforeman.org/scripts/test/test_hammer_cli_foreman_pull_request.sh#L17

Tests pass locally, but fail in Jenkins:

```
20:33:10 Bundler could not find compatible versions for gem "hammer_cli":
20:33:10   In Gemfile:
20:33:10     hammer_cli
20:33:10 
20:33:10     hammer_cli_foreman was resolved to 0.15.2, which depends on
20:33:10       hammer_cli (>= 0.15.1, < 0.16.0)
20:33:10 
20:33:10 Could not find gem 'hammer_cli (>= 0.15.1, < 0.16.0)', which is required by gem
20:33:10 'hammer_cli_foreman', in any of the relevant sources:
20:33:10   git://github.com/theforeman/hammer-cli.git (at master@107a512)
```

mbacovsky:
@jturel thanks for the patch. I guess you are right with the cause. It would make more sense to test against master only if the PR target is master and let bundler install hammer-cli from gems otherwise. Are you familiar with Jenkins job builder scripts?

jturel:
@mbacovsky this might do the trick: https://github.com/theforeman/foreman-infra/pull/1026

mbacovsky:
[test hammer]

mbacovsky:
[hammer test]

mbacovsky:
[test hammer]

mbacovsky:
@jturel new gem was pushed to rg.o. Thanks for your help!

jturel:
[test hammer]

PR #420, opened 2019-04-24
Fixes #26668 - adds project domain flags

lzap:
For more info see https://github.com/theforeman/foreman/pull/6706

### DIFF
--- a/lib/hammer_cli_foreman/compute_resource.rb
+++ b/lib/hammer_cli_foreman/compute_resource.rb
@@ -43,7 +43,9 @@ module HammerCLIForeman
           Fields::Field.new(:label => _('Server'), :path => [:server])
         ],
         'openstack' => [
-          Fields::Field.new(:label => _('Tenant'), :path => [:tenant])
+          Fields::Field.new(:label => _('Tenant'), :path => [:tenant]),
+          Fields::Field.new(:label => _('Project domain name'), :path => [:project_domain_name]),
+          Fields::Field.new(:label => _('Project domain ID'), :path => [:project_domain_id])
         ],
         'rackspace' => [
           Fields::Field.new(:label => _('Region'), :path => [:region])


mbacovsky:
:+1: Thanks @lzap! Looks good. I'll test once the core PR is merged.

ofedoren:
I guess, it'll be also quite good to update the help for OpenStack provider, so the user knows about the new parameters (e.g. for host creation).

lzap:
Merged!

lib/hammer_cli_foreman/compute_resource.rb
```diff
@@ -43,7 +43,9 @@ class InfoCommand < HammerCLIForeman::InfoCommand
           Fields::Field.new(:label => _('Server'), :path => [:server])
         ],
         'openstack' => [
-          Fields::Field.new(:label => _('Tenant'), :path => [:tenant])
+          Fields::Field.new(:label => _('Tenant'), :path => [:tenant]),
+          Fields::Field.new(:label => _('Project domain name'), :path => [:project_domain_name]),
+          Fields::Field.new(:label => _('Project domain ID'), :path => [:project_domain_id])
```
shiramax: thanks @lzap, Looks good :)
If you are already changing this code, I think It would be better to have in each compute resource file, a function for 'provider_spesific_fileds'  so that compute resource plugins will able to add their specific details easily

for example: 

**/hammer-cli-foreman/lib/hammer_cli_foreman/compute_resource/base.rb**
will contain:
`def provider_specific_fields; []; end
`
**/hammer-cli-foreman/lib/hammer_cli_foreman/compute_resource/openstack.rb** 
will contain:

```
def provider_spesific_fileds
   [
          Fields::Field.new(:label => _('Tenant'), :path => [:tenant]),
          Fields::Field.new(:label => _('Project domain name'), :path => [:project_domain_name]),
          Fields::Field.new(:label => _('Project domain ID'), :path => [:project_domain_id])
    ]
end 
```


and the print data function in **compute_resource.rb** will look like that:

```
def print_data(data)
        provider = ::HammerCLIForeman.compute_resources[data["provider"].downcase]
        if provider
          output_definition.fields.concat(provider.provider_specific_fields || [])
          super(data)
        end
 end
```

kgaikwad: @lzap, @shiramax,

Added suggested changes into [PR-431](https://github.com/theforeman/hammer-cli-foreman/pull/431)

lzap:
Sorry for the delay, @mbacovsky has more expertise than me on that but overall looks like a good idea to me.

mbacovsky:
I like the approach @shiramax suggested. @lzap would you mind fixing the conflicts once https://github.com/theforeman/hammer-cli-foreman/pull/431 is merged?

mbacovsky:
I've just seen the core PR related to #431 is not merged yet so merging this as is and letting @kgaikwad fix the conflicts would be also an option. 
@shiramax, as you are actively watching all three PRs what would your preferred way to get these in?

lzap:
I am fine with fixing conflicts or whatever is needed, please decide guys.

mbacovsky:
:+1: Lets merge it now, thanks @lzap, @shiramax, @kgaikwad!

PR #424, opened 2019-06-20
Fixes #25093, #26421 - Host creation with multi SCSI controllers

ofedoren:
The problem is that help is not quite usefull for creating a host with multiple SCSI controllers. Server's API accepts several JSON strings and none of them is described.

My suggestion is to send JSON (e.g. compute attributes) similar to JSON which is being created from UI.

This PR contains a small help fix as well as refactoring of compute resource help builder to be more general. If there is no point in that (or it's unreadable), I can remove it and fix help entries only.


ofedoren:
@ezr-ondrej, you might be interested in it as you created one of issues.

If you've got an environment with working VMware provider, can you confirm that a command with the following structure will create a host with multiple SCSI controllers? CHANGEME should be replaced with your specific data.
```bash
hammer host create \
  --location-id CHANGEME \
  --organization-id CHANGEME \
  --name CHANGEME \
  --compute-attributes '{"cpus":1,"corespersocket":1,"memory_mb":2048,"firmware":"bios","resource_pool":"CHANGEME","cluster":"CHANGEME","guest_id":"CHANGEME","path":"CHANGEME","hardware_version":"Default","memoryHotAddEnabled":"0","cpuHotAddEnabled":"0","add_cdrom":"0","start":"0","scsi_controllers":[{"type":"ParaVirtualSCSIController","key":1000},{"type":"ParaVirtualSCSIController","key":1001}]}' \
  --interface '{"type":"Nic::Managed","name":"CHANGEME","domain_id":CHANGEME,"managed":true,"primary":true,"provision":true,"virtual":false,"compute_attributes":{"type":"VirtualE1000","network":"CHANGEME"}}' \
  --domain CHANGEME \
  --operatingsystem CHANGEME\
  --root-password CHANGEME \
  --medium CHANGEME \
  --mac CHANGEME \
  --architecture CHANGEME \
  --partition-table-id CHANGEME \
  --volume '{"datastore":"CHANGEME","name":"Hard disk 00","size_gb":"10","thin":"true","eager_zero":"false","controller_key":1000}' \
  --volume '{"datastore":"CHANGEME","name":"Hard disk 01","size_gb":"10","thin":"true","eager_zero":"false","controller_key":1000}' \
  --volume '{"datastore":"CHANGEME","name":"Hard disk 10","size_gb":"10","thin":"true","eager_zero":"false","controller_key":1001}' \
  --volume '{"datastore":"CHANGEME","name":"Hard disk 11","size_gb":"10","thin":"true","eager_zero":"false","controller_key":1001}'
```

lib/hammer_cli_foreman.rb
```diff
@@ -3,6 +3,15 @@
 require 'apipie_bindings'
 
 module HammerCLIForeman
+  @compute_resources = {}
```
shiramax: why are you moving in from lib/hammer_cli_foreman/compute_resource/register_compute_resources.rb ?
there are plugins that depend on these methods, and this is not so good to change those

lib/hammer_cli_foreman/compute_resources/ec2.rb
```diff
@@ -0,0 +1,19 @@
+module HammerCLIForeman
+  module ComputeResources
+    class EC2 < Base
```
shiramax: please don't change the path of this files, 
there are plugins that depend on these paths, and this is not so good to change those (for example hammer_cli_foreman_kubevirt)

ofedoren:
@shiramax, thanks! I've returned pathes you mentioned.

ofedoren:
@shiramax, updated.

Thanks, I forgot to update compute-attributes module :/

BTW, what do you think about small changes in help? Is it clear enough what should be used to create multiple SCSI controllers?

lib/hammer_cli_foreman/command_extensions/hosts/help/compute_resources.rb
```diff
@@ -0,0 +1,91 @@
...
+              new_self.class_eval(help_block)
+              new_self.customized = true
+              new_self
+            end
+
+            def help_block
+              help_block = <<HELP_BLOCK
+                help do |h|
+                  h.section(_('Provider specific options')) do |h|
+                    HammerCLIForeman.compute_resources.each do |name, provider|
```
shiramax: I don't feel that putting this code under block is the right approach, I feel like it's not readable like that, what was the reason for changing this code?

lib/hammer_cli_foreman/command_extensions/hosts/help/compute_resources.rb
```diff
@@ -0,0 +1,91 @@
...
+                        end
+                        if %i[all volume].include?(attributes)
+                          h.section('--volume') do |h|
+                            h.list(provider.volume_attributes)
+                          end
+                        end
+                        if %i[all interface].include?(attributes)
+                          h.section('--interface') do |h|
+                            if host_interfaces
+                              h.list(host_interfaces)
```
shiramax: I don't feel that putting this code under block is the right approach, I feel like it's not readable like that, what was the reason for changing this code?

lib/hammer_cli_foreman/compute_resource/vmware.rb
```diff
@@ -1,68 +1,79 @@
...
+          ['memoryHotAddEnabled',  _('Must be a 1 or 0, lets you add CPU resources while the machine is on')],
+          ['cpuHotAddEnabled',     _('Must be a 1 or 0, lets you add memory resources while the machine is on')],
+          ['add_cdrom',            _('Must be a 1 or 0, Add a CD-ROM drive to the virtual machine')],
+          ['start',                _('Must be a 1 or 0, whether to start the machine or not')],
+          ['annotation',           _('Annotation Notes')],
+          ['scsi_controllers',     _('List with SCSI controllers definitions')]
         ]
       end
 
-      def host_attributes
```
shiramax: start a VM is an action that is not part of the compute attributes (when creating a compute profile, that's why we had the host_attributes

lib/hammer_cli_foreman.rb
```diff
@@ -22,6 +22,7 @@ def self.exception_handler_class
   require 'hammer_cli_foreman/logger'
 
   begin
+    require 'hammer_cli_foreman/compute_resource/register_compute_resources'
```
shiramax: I'm not sure we need to register the compute resources in this phase. It's only needed in host and  compute-profile command

lib/hammer_cli_foreman/command_extensions/hosts/help/compute_resources.rb
ofedoren: To be able create custom help extensions, see: 
https://github.com/theforeman/hammer-cli-foreman/blob/97d96fec40e50a82de12192259fa19c6c96d7a6e/lib/hammer_cli_foreman/compute_attribute.rb#L137

https://github.com/theforeman/hammer-cli-foreman/blob/97d96fec40e50a82de12192259fa19c6c96d7a6e/lib/hammer_cli_foreman/host.rb#L304

lib/hammer_cli_foreman/compute_resource/vmware.rb
ofedoren: You're right, returned to the previous state.

lib/hammer_cli_foreman.rb
ofedoren: Returned to the previous state.

ofedoren:
@shiramax, updated.

lib/hammer_cli_foreman/compute_resource/libvirt.rb
```diff
@@ -2,48 +2,48 @@ module HammerCLIForeman
...
       end
 
       def compute_attributes
         [
-            ['cpus',   _('Number of CPUs')],
-            ['memory', _('String, amount of memory, value in bytes')],
-            ['start',  _('Boolean (expressed as 0 or 1), whether to start the machine or not')]
+          ['cpus',   _('Number of CPUs')],
+          ['memory', _('String, amount of memory, value in bytes')],
+          ['start',  _('Boolean (expressed as 0 or 1), whether to start the machine or not')]
```
shiramax: this line is not necessary, you already have 'start' in host_attributes

ofedoren: Since it was there before this PR, I missed this one, thanks :)

ezr-ondrej:
Needed to change the attributes of `interface` to their new format values `compute-type` and `compute-network`, but than it works smoothly.

`--start` parameter is in `--interface` help section though, could we do something about it?
```shell
    --interface:
      compute_type        Type of the network adapter, for example one of:
                            VirtualVmxnet3
                            VirtualE1000
                          See documentation center for your version of vSphere to find more details about available adapter types:
                            https://www.vmware.com/support/pubs/
      compute_network     Network ID from VMware
      start               Must be a 1 or 0, whether to start the machine or not
```

ofedoren:
@shiramax, updated.

So I assume --start should be in --compute-attributes section in the host creation help? It is now, but if it's wrong, poke me and I'll change it.

ezr-ondrej:
@ofedoren it is wrong :) it should be a common vm parameter :)

shiramax:
@ofedoren , @ezr-ondrej AFAIK the start should be in --compute-attributes section only in the host creation help. 
@ezr-ondrej what do you mean by common name parameter?

ofedoren:
@ezr-ondrej, I've noticed that `start` is going under `compute_attributes` when I looked how does UI set the parameters in `@host` object.

ezr-ondrej:
> @ofedoren , @ezr-ondrej AFAIK the start should be in --compute-attributes section only in the host creation help.

Sorry @ofedoren, misread you! in the common vm I meant in `--compute-attributes` section, so the same you said. My bad :man_facepalming:

So is it ready to retest?

ofedoren:
@ezr-ondrej, no problem :) Yes, it is.

ezr-ondrej:
Works well :+1:
The `--start` is still in the `--interface` section for vmware though.

ofedoren:
@ezr-ondrej, are you sure? I force pushed the changes, maybe you haven't got the new version?

Because I see it under --compute-attrs section:
ScreenShot-1563282382268

ezr-ondrej:
Yeah, I was sure, I updated, didn't checked out, though :face_with_thermometer:
Retested, works and the `--start` is where it should be.

PR #429, opened 2019-07-18
Fixes #27339 - Report template schedule works with --name

ofedoren:
When a user specifies `--name` option it is automatically resolved into `id` and saved into `options` under `option_id` key. The `option_id` method can be used only when a user has explicitly specified it. That caused the issue.


shiramax:
@ofedoren could you please add tests?

ofedoren:
@shiramax, updated :)

PR #432, opened 2019-07-22
Fixes #27343 - consider value not display name of compute_resource

kgaikwad:
adding @shiramax and @ezr-ondrej into the loop

Provider friendly name of GCE compute resource is `Google` so compute attribute flow is broken for GCE.

### DIFF
--- a/lib/hammer_cli_foreman/compute_attribute.rb
+++ b/lib/hammer_cli_foreman/compute_attribute.rb
@@ -1,5 +1,6 @@
-require 'hammer_cli_foreman/image'
 require 'hammer_cli_foreman/compute_resource/register_compute_resources'
+require 'hammer_cli_foreman/compute_resource/utils'
+require 'hammer_cli_foreman/image'
 
 module HammerCLIForeman
   class ComputeAttribute < HammerCLIForeman::Command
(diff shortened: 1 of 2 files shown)


lib/hammer_cli_foreman/compute_attribute.rb
```diff
@@ -37,7 +37,7 @@ class Create < HammerCLIForeman::CreateCommand
 
       def request_params
         params = super
-        compute_resource_name = HammerCLIForeman.record_to_common_format(HammerCLIForeman.foreman_resource(:compute_resources).call(:show, 'id' => options["option_compute_resource_id"] ))["provider_friendly_name"].downcase
+        compute_resource_name = HammerCLIForeman.record_to_common_format(HammerCLIForeman.foreman_resource(:compute_resources).call(:show, 'id' => options["option_compute_resource_id"] ))["provider"].downcase
```
mbacovsky: All those identical lines are asking for DRYing. Could we have something like:
```ruby
compute_resource_name = HammerCLIForeman.ComputeResources.resource_provider(options["option_compute_resource_id"])
```
I'd expect the function to live in `hammer_cli_foreman/compute_resource/utils.rb`

kgaikwad: @mbacovsky, 
To resolve this I need to require `hammer_cli_foreman/compute_resource/utils` file.

I am not sure that how this [method call](https://github.com/theforeman/hammer-cli-foreman/blob/master/lib/hammer_cli_foreman/hosts/common_update_options.rb#L86) is being working.
When I debugged further using pry on host update then I wasn't able to see method `get_host_compute_resource_id` on `::HammerCLIForeman::ComputeResources`. Am I missing something?

mbacovsky: @kgaikwad you are right, we miss require on the cr utils file. We've lost it two weeks ago in here https://github.com/theforeman/hammer-cli-foreman/commit/7b59a148f508e314b1817930a94b0c989e803c35#diff-bbed5e1ba9432426e1430eeca619f8ba :(

kgaikwad: Do you want me to add this require statement for `host` in this PR or separately?

ofedoren: There is an [issue](https://projects.theforeman.org/issues/27444), so that fix will be in a separate PR.

kgaikwad: Updated PR as per @mbacovsky suggested and created one #434 for host.

PR #434, opened 2019-07-25
Fixes #27444 - fix missing helper methods for host

kgaikwad:


mbacovsky:
@kgaikwad I really appreciate your effort! There was also PR #435 proposed for the same issue. Even though it landed here later then yours I merged the #435 because it has also tests that help us unsure we don't lose the utils again.

PR #439, opened 2019-08-19
Fixes #27652 - Fix interfaces when creating a host

shiramax:
Creating a host with interfaces didn't work properly. 
the reason for that is that in the compute profile PR, I removed the "compute_" string from each attribute in the interfaces, that worked for the compute profile but didn't work for the host creation. 
the reason for that is this functiom: 
```
 def interfaces_attributes
        return [] if option_interface_list.empty?

        # move each attribute starting with "compute_" to compute_attributes
        option_interface_list.collect do |nic|
          compute_attributes = {}
          nic.keys.each do |key|
            if key.start_with? 'compute_'
              compute_attributes[key.gsub('compute_', '')] = nic.delete(key)
            end
          end
          subnet_name = nic.delete('subnet')
          nic['subnet_id'] ||= subnet_id(subnet_name) if subnet_name
          domain_name = nic.delete('domain')
          nic['domain_id'] ||= domain_id(domain_name) if domain_name
          nic['compute_attributes'] = compute_attributes unless compute_attributes.empty?
          nic
        end
      end
```
that puts the interfaces attributes (that are spesfic to a compute resource) inside "compute_attributes"
the function  checks if the attribute starts with "compute" and then puts it under "compute_attributes" 

so I needed to restore the "compute_" string and add a function in compute_attribute that removes it.

### DIFF
--- a/lib/hammer_cli_foreman/compute_attribute.rb
+++ b/lib/hammer_cli_foreman/compute_attribute.rb
@@ -21,6 +21,22 @@ module HammerCLIForeman
       attribute_list.size.times.map { |idx| idx.to_s }.zip(attribute_list).to_h
     end
 
+    def self.alter_interface(interface)
+      # move each attribute starting with "compute_" to compute_attributes
+        interface.keys.each do |key|
+          if key.start_with? 'compute_'
+            interface[key.gsub('compute_', '')] = interface.delete(key)
+          end
+        end
+        interface
+    end
+
+    def self.alter_interfaces_list(interfaces_list)
+      interfaces_list.collect do |nic|
+        alter_interface(nic)
+      end
+    end
+
     class Create < HammerCLIForeman::CreateCommand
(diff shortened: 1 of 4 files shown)


mbacovsky:
@shiramax, the code looks reasonable I need to do some testing though.
My only concern so far is interface_attributes in VmWare. They were prefixed before and with this patch the 'compute_' will be stripped down, correct? Will it still work?

shiramax:
@mbacovsky I think that VMWare wasn't tested  before, But see :

mbacovsky:
:+1: @shiramax it worked for me for libvirt. I was able to create interfaces in compute profile as well as in host.

@ofedoren don't you have env to test this for vmware?

ofedoren:
@mbacovsky, unfortunatelly I don't have it currenlty. But I hope @ezr-ondrej can help with this one.

PR #442, opened 2019-09-02
Refs #17819 - Allow adapters print page by page

ofedoren:
This PR shows how to use https://github.com/theforeman/hammer-cli/pull/311 feature.

For testing purposes you might want to change https://github.com/theforeman/hammer-cli-foreman/blob/2f3b422cacce83e5c4e7b0544b7fedd92c509a70/lib/hammer_cli_foreman/commands.rb#L230 1000 to lower value (e.g. 5 or 10).

### DIFF
--- a/lib/hammer_cli_foreman/commands.rb
+++ b/lib/hammer_cli_foreman/commands.rb
@@ -185,7 +185,9 @@ module HammerCLIForeman
     end
 
     def send_request
-      transform_format(super)
+      data = super
+      @meta = retrieve_meta(data)
+      transform_format(data)
     end
 
     def transform_format(data)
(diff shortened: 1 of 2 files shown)


xprazak2:
One small drawback is that column width for table is calculated on each chunk, which results in different column widths for each chunk:

chunks

But I do not think it matters that much since the chunk size is 1000. Works flawlessly for json and yaml.

ofedoren:
> One small drawback is that column width for table is calculated on each chunk, which results in different column widths for each chunk:
> 
> chunks
> 
> But I do not think it matters that much since the chunk size is 1000. Works flawlessly for json and yaml.

yeap, I was thinking about that as well and was thinking how to avoid that, but since the width is calculated for each chunk when it's available, there is no way how to determine the largest possible incoming chunk. The one of the solutions would be to set the default width to some value, but I don't think it's a good way...

xprazak2:
[test hammer]

:+1: from me, lets see if the tests are green.

ofedoren:
@xprazak2, unfortunatelly they will not be green until the required PR is merged :(

xprazak2:
[test hammer]

xprazak2:
Thanks!

PR #448, opened 2019-09-20
Fixes #27868 - Add option to support host's param type

ofedoren:
Requires https://github.com/theforeman/hammer-cli/pull/316

Or we could remove `:host_parameters_attributes` from https://github.com/theforeman/hammer-cli-foreman/blob/9999e05ae70fe68ccfe8fab76c70349dfefe1e45/lib/hammer_cli_foreman/hosts/common_update_options.rb#L45 to build the option from apidocs and thus the core changes will be not required.

Note: if both `--parameters` and `--typed-parameters` are specified then the values from `--parameters` are not overrided by values from `--typed-parameters`.

### DIFF
--- a/lib/hammer_cli_foreman/hosts/common_update_options.rb
+++ b/lib/hammer_cli_foreman/hosts/common_update_options.rb
@@ -30,6 +30,12 @@ module HammerCLIForeman
 
         base.option "--parameters", "PARAMS", _("Replaces with new host parameters"),
           :format => HammerCLI::Options::Normalizers::KeyValueList.new
+
+        host_params = base.resource.action(base.action).params
+                          .find { |p| p.name == 'host' }.params
+                          .find { |p| p.name == 'host_parameters_attributes' }
+        base.option "--typed-parameters", "TYPED_PARAMS", _("Replaces with new host parameters (with type support)"),
+          :format => HammerCLI::Options::Normalizers::ListNested.new(host_params.params)
         base.option "--compute-attributes", "COMPUTE_ATTRS", _("Compute resource attributes"),
           :format => HammerCLI::Options::Normalizers::KeyValueList.new
         base.option "--volume", "VOLUME", _("Volume parameters"), :multivalued => true,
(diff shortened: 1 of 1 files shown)


ofedoren:
@shiramax, thanks for review! :)

I'm adding a new option because of parsing the values provided for that option(s).

For example option `--parameters` has a key=value normalizer, so it accepts a list of key=value only:
```sh
hammer host update --parameters "par1=val1,par2=val2"
```
But option `--typed-parameters` has a bit more complex normalizer which allows to pass a bit more comlex value defined by a schema:
```sh
hammer host update --typed-parameters "name=par1\,value=val1\,parameter_type=string\,hidden_value=false,name=par2\,value=val2\,parameter_type=string\,hidden_value=false"
```

Please notice that we cannot pass the parameter_type for `--parameters` option, but we can for `--typed-parameters`. Also, we cannot just wipe `--parameters` option or change it to have normalizer like `--typed-parameters` does since it could break automatization used in scripts.

shiramax:
Thanks for the explanation @ofedoren, I think this input of the typed-parameter, can be a bit confusing: "name=par1\,value=val1\,parameter_type=string\,hidden_value=false,name=par2\,value=val2\,parameter_type=string\,hidden_value=false"
"

I wonder if we could to do a different format for this, for example, maybe we could do this like we do the --interfaces or --volume. it means --typed-parameters should be a parameter we could speficy multiple times. also we will need to generate help for the parameters, and finally you should change the parameters that send to the server. 
It might be more work, but I think it would be more clear how to use the method.

mbacovsky:
I quite like the solution that @shiramax suggested. That would be probably easier to input manually for the users. For the solution provided in this PR I can see benefits for automation where the parameters are added in scripts especially in JSON format.

It seems these two ways are not conflicting and there is chance to add also the multivalue variant @shiramax suggested. I assume the name of the option would be --typed-parameter (singular) so there is no conflict. It may be confusing if it is not consistent though.
 @shiramax sorry for merging without your approval but it seems your questions were answered and this needs to be part of 1.24. If you have any concerns, please let me know and we can resolve this in minor release.

PR #453, opened 2019-11-20
Fixes #28196 - handle error when empty parameters passed

rahulbajaj0509:


ofedoren:
@rahulbajaj0509, @mbacovsky, LGTM as well, but don't we need to do the same for https://github.com/theforeman/hammer-cli-foreman/blob/097de65123949dbd5dac9c524094387fbe6a3506/lib/hammer_cli_foreman/api/oauth/authentication_code_grant.rb#L35 ?

PR #458, opened 2019-12-02
Fixes #28098 - add report-template import command

yifatmakias:


test/functional/report_template_test.rb
```diff
@@ -129,6 +129,35 @@
     end
   end
 
+  describe 'import' do
+    let(:cmd) { %w(report-template import) }
+    let(:tempfile) { Tempfile.new('template') }
+    it 'requires --name or --file' do
```
ofedoren: `requires --name AND --file` makes more sense, since they are both requied.

yifatmakias: True, I will fix it.

PR #462, opened 2019-12-09
Fixes #28454 - added disk interface to ovirt vm creation help

yifatmakias:


lib/hammer_cli_foreman/compute_resource/ovirt.rb
```diff
@@ -35,7 +35,8 @@ def volume_attributes
           ['storage_domain', _('ID of storage domain')],
           ['bootable',       _('Boolean, only one volume can be bootable')],
           ['preallocate',    _('Boolean, set %s to preallocate') % 'true'],
-          ['wipe_after_delete', _('Boolean, set %s to wipe disk after delete') % 'true']
+          ['wipe_after_delete', _('Boolean, set %s to wipe disk after delete') % 'true'],
```
shiramax: AFAIK there could be only three types of disk interfaces, so please let the user know all his possibilities.

yifatmakias: Ok, fixed it, see if it is ok now :)

PR #466, opened 2019-12-12
Refs #28135 - Remove URL field from GCE info

ofedoren:
This should be also cherry-picked into 0.19-stable for 0.19.5.

### DIFF
--- a/lib/hammer_cli_foreman/compute_resource.rb
+++ b/lib/hammer_cli_foreman/compute_resource.rb
@@ -31,7 +31,6 @@ module HammerCLIForeman
       include ProviderNameLegacy
 
       output ListCommand.output_definition do
-        field :url, _("Url")
         field :description, _("Description")
         field :user, _("User")
         HammerCLIForeman::References.taxonomies(self)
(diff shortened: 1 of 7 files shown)


xprazak2:
Before:

url-before

After: 

url-after

I haven't tested vmware, ec2, ovirt and rackspace - but openstack and libvirt show url correctly and given they all use the same code, I do not see a reason why the rest of the compute resources would not show the field properly.
I think this is ready unless someone wants to test those remaining CRs.

ofedoren:
@shiramax, I think we could stop including `url` attribute in JSON responses from API for GCE info, but it isn't backward compatible, is it?

shiramax:
@ofedoren OK, I checked with foreman, and it seems like this is the best option. thanks @ofedoren !

kgaikwad:
@ofedoren, @shiramax, 
any updates on this? can we merge this PR?

ofedoren:
@kgaikwad, sure, if nobody has concerns, it's ready.

kgaikwad:
@shiramax, This change should be also cherry-picked into 0.19-stable.

ofedoren:
@kgaikwad, I think we need a CP for this into 0.19-stable. Do you want to do it or should I?

kgaikwad:
@ofedoren, 
Either way will work for me. As I didn't see any CP PR so asked. It would be good if you do it as original PR created by you.

PR #468, opened 2019-12-13
Fixes #28503 - Increase ID column width for available networks

apuntamb:
The available_networks call for the azure compute resource lists the id and name, but the id is too long and hence, gets trimmed that blocks the user to go ahead and provide interfaces_attributes while creating a host. The PR handles that width of the ID column.

### DIFF
--- a/lib/hammer_cli_foreman/compute_resource.rb
+++ b/lib/hammer_cli_foreman/compute_resource.rb
@@ -121,7 +121,7 @@ module HammerCLIForeman
       command_name 'networks'
 
       output do
-        field :id, _('Id')
+        field :id, _('Id'), Fields::Field, :max_width => 200, :hide_blank => true
         field :name, _('Name')
       end
 


apuntamb:
@shiramax @ofedoren I'd tested this and it worked for me.

ofedoren:
@apuntamb, don't you think that 200 chars is quite a lot for column width?..

apuntamb:
> @apuntamb, don't you think that 200 chars is quite a lot for column width?..
> 
> Also, I think that for `Id` column a simple `Fields::Field` is enough. Although `Fields::List` is the same, it's less readable in this context.

@ofedoren I agree 200 is quite a long width especially for an ID column. But, for azure, the ID string is really long.
Example of the longest ID -
`/subscriptions/11111111-1111-1111-1111-111111111111/resourceGroups/aaaabbbbbcccc12/providers/Microsoft.Network/virtualNetworks/ddddeeeeeffff12gggggg1Vnet/subnets/hhhhiiiiijjjj12kkkkkk1Subnet`
How about if we could continue the string in the next line?
Thoughts?

ofedoren:
@apuntamb, since the hammer's table output is quite limited I've got two solutions:

 - I'll make a PR to hammer core that will allow the columns have :max_width more than 80 characters (then you can set in this PR :max_width for `Id` field to 200 or whatever)
 - We leave this as it is, but then to see the full ID a user will need to change the output adapter (e.g. specify `--output=base` option. The `base` adapter doesn't cut any data to fill the column.

Both ways are OK in my opinion, because if we go the first way, there is a possibility that printed data aren't readable that much, see:

With `max_width: 200` and my default terminal font size:
ScreenShot-1576495170332
But with `max_width: 200` and slightly smaller font size it's ok:
ScreenShot-1576495253836

So again in my opinion it always goes back to user: either change the output adapter or font size.

Please tell me your opinion and if you prefer the first way, I'll create the PR.

apuntamb:
@ofedoren I think we can go with specifying `--output=base` here. As I can see the output is:
```
Id:  /subscriptions/11111111-1111-1111-1111-111111111111/resourceGroups/aaaabbbbbcccc12/providers/Microsoft.Network/virtualNetworks/ddddeeeeeffff12gggggg1Vnet/subnets/hhhhiiiiijjjj12kkkkkk1Subnet
Name: hhhhiiiiijjjj12kkkkkk1Subnet
```
which doesn't look so bad. So, I think changing the output adapter might be a better option here. So, do I need to make any code changes in that case?

ofedoren:
@apuntamb, if we go the second way (just specify `--output=base` and get the result as you've got it) then this PR is not needed at all :(

Specifying `:width` for field is just needed for `--output=table` which is default for `List` commands. If we use `--output=base` (which is default for `Info` commands) then we don't need to specify `:width` as it is ignored by this output adapter.

apuntamb:
@ofedoren Okk makes sense. Just before closing the PR then, I would like to know if we could show a Warning or just a help text for that matter for the user to understand that the command best works with `--ouput=base` in case of azure. If so, 1) do we need to add that message in hammer? or, 2) do we need to add it in `#available_networks` method in the plugin?

lib/hammer_cli_foreman/compute_resource.rb
```diff
@@ -121,7 +121,7 @@ class AvailableNetworksCommand < HammerCLIForeman::ListCommand
       command_name 'networks'
 
       output do
-        field :id, _('Id')
+        field :id, _('Id'), Fields::List, :width => 200, :hide_blank => true
```
ofedoren: For this field `Fields::Field` is enough and more readable in my opinion.

And if we want the field to not be static (200 chars only), but squeezed or expanded according to the data (e.g. 20 chars data => 20 chars width of the column), please change the `:width` option to `:max_width`.

apuntamb: @ofedoren Made changes as per suggestions. Let me know if I missed something.

shiramax:
thanks, @apuntamb, and @ofedoren !

PR #469, opened 2019-12-19
Fixes #20451 - added virtual machine command for compute resource

yifatmakias:
This PR is related to this foreman PR:
https://github.com/theforeman/foreman/pull/7276

### DIFF
--- a/lib/hammer_cli_foreman/virtual_machine.rb
+++ b/lib/hammer_cli_foreman/virtual_machine.rb
@@ -1,0 +1,55 @@
+
+module HammerCLIForeman
+
+  class VirtualMachine < HammerCLIForeman::Command
+    resource :compute_resources
+    command_name 'virtual_machine'
+    desc _("View and manage compute resource's virtual machines")
+
+    class PowerVmCommand < HammerCLIForeman::Command
+      action :power_vm
+      command_name 'power'
+      option "--vm-id", "VM-ID", _("Virtual machine id, for gce use virtual machine name")
+
+      success_message _("Virtual machine is powering.")
+      failure_message _("Could not power the virtual machine")
+
+      build_options
+    end
+
+    class InfoCommand < HammerCLIForeman::InfoCommand
+      action :show_vm
+      option "--vm-id", "VM-ID", _("Virtual machine id, for gce use virtual machine name")
+
+      output do
+        field :id, _("Id")
+        field :name, _("Name")
+      end
+
+      def print_data(data)
+        provider = ::HammerCLIForeman.compute_resources[data['provider'].downcase]
+        if provider
+          output_definition.fields.concat(provider.provider_vm_specific_fields || [])
+          super(data)
+        end
+      end
+
+      build_options
+    end
+
+    class DeleteCommand < HammerCLIForeman::DeleteCommand
+      action :destroy_vm
+      option "--vm-id", "VM-ID", _("Virtual machine id, for gce use virtual machine name")
+
+
+      success_message _("Virtual machine deleted.")
+      failure_message _("Could not delete the virtual machine")
+
+      build_options
+    end
+
+    autoload_subcommands
+  end
+
+end
+
--- a/test/functional/virtual_machine_test.rb
+++ b/test/functional/virtual_machine_test.rb
@@ -1,0 +1,129 @@
+
+require File.join(File.dirname(__FILE__), 'test_helper')
+
+require 'hammer_cli_foreman/compute_resource'
+
+describe HammerCLIForeman::VirtualMachine do
+
+  context "InfoCommand" do
+    before do
+      @cmd = ["compute-resource", "virtual-machine", "info"]
+      @vm = {
+          'id' => 1,
+          'name' => 'vm1',
+          'provider' => 'Ovirt',
+          'cpu' => 1,
+          'memory' => 1,
+          'status' => 'down'
+      }
+    end
+
+    it "should print error on missing --id and --vm-id" do
+      api_expects_no_call
+      result = run_cmd(@cmd)
+      assert_match("Missing arguments for '--id', '--vm-id'.\n", result.err)
+    end
+
+    it "should print error on missing --vm-id" do
+      params = ['--id=1']
+
+      api_expects_no_call
+      result = run_cmd(@cmd + params)
+      assert_match("Missing arguments for '--vm-id'.\n", result.err)
+    end
+
+    it "should return vm info" do
+      params = ['--id=1', '--vm-id=1']
+      api_expects(:compute_resources, :show_vm) do |par|
+        par['id'] == '1' && par['vm_id'] == '1'
+      end.returns(@vm)
+      result = run_cmd(@cmd + params)
+      assert_match("Id:     1\nName:   vm1\nCPUs:   1\nMemory: 1\nStatus: down\n\n", result.out)
+    end
+
+  end
+
+  context "ListCommand" do
+    before do
+      @cmd = ["compute-resource", "virtual-machines"]
+      @available_virtual_machines = [
+        { id: 1, name: 'vm1' },
+        { id: 2, name: 'vm2' }
+      ]
+    end
+
+    it "should print error on missing --id" do
+      api_expects_no_call
+      result = run_cmd(@cmd)
+      assert_match("Missing arguments for '--id'.\n", result.err)
+    end
+
+    it "should return vms list" do
+      params = ['--id=1']
+      api_expects(:compute_resources, :available_virtual_machines) do |par|
+        par['id'] == '1'
+      end.returns(@available_virtual_machines)
+      result = run_cmd(@cmd + params)
+      assert_match(/-----|---\nNAME | ID\n-----|---\nvm1  | 1 \nvm2  | 2 \n-----|---\n/, result.out)
+    end
+  end
+
+  context "PowerCommand" do
+    before do
+      @cmd = ["compute-resource", "virtual-machine", "power"]
+    end
+
+    it "should print error on missing --id and --vm-id" do
+      api_expects_no_call
+      result = run_cmd(@cmd)
+      assert_match("Could not power the virtual machine:\n  Missing arguments for '--id', '--vm-id'.\n", result.err)
+    end
+
+    it "should print error on missing --vm-id" do
+      params = ['--id=1']
+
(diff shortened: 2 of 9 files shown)


yifatmakias:
This PR is now related also to this pr:
https://github.com/theforeman/hammer-cli-foreman/pull/473

the tests will not pass until the second pr will be merged because of a change in the foreman-API.

ofedoren:
[test hammer]

yifatmakias:
This pr is depending on the following pr:
https://github.com/theforeman/hammer-cli-foreman/pull/485

That is because the foreman_api.json should be changed (its name is not correct) and then we could change the test API version to 2.0 (in the test_helper.rb file) and all the tests will pass.

xprazak2:
Setting the label as per [previous](https://github.com/theforeman/hammer-cli-foreman/pull/469#issuecomment-574190583) comment.

yifatmakias:
@xprazak2 It is no longer blocked by another PR, so the label can be removed now. :)

lib/hammer_cli_foreman/virtual_machine.rb
```diff
@@ -0,0 +1,78 @@
...
+
+  class VirtualMachine < HammerCLIForeman::Command
+
+    resource :compute_resources
+    command_name 'virtual_machine'
+    desc _("View and manage compute resource's virtual machines")
+
+
+    class AvailableVirtualMachinesCommand < HammerCLIForeman::ListCommand
+
```
ofedoren: Redundant spaces between rows.  Applies for command definitions below as well.

lib/hammer_cli_foreman/virtual_machine.rb
```diff
@@ -0,0 +1,78 @@
...
+      resource :compute_resources, :available_virtual_machines
```
ofedoren: If `resource` was specified in parent command/class, there is no need to specify it again.

Simple `action :available_virtual_machines` is enough. Applies for command definitions below as well.

lib/hammer_cli_foreman/virtual_machine.rb
```diff
@@ -0,0 +1,78 @@
...
+module HammerCLIForeman
+
+  class VirtualMachine < HammerCLIForeman::Command
+
+    resource :compute_resources
+    command_name 'virtual_machine'
+    desc _("View and manage compute resource's virtual machines")
+
+
+    class AvailableVirtualMachinesCommand < HammerCLIForeman::ListCommand
```
ofedoren: I'd actually move this command here:

https://github.com/theforeman/hammer-cli-foreman/blob/f163dd9fc9a232f711621f81a16752624edd5de7/lib/hammer_cli_foreman/compute_resource.rb#L213-L226

since this follows the same pattern.

lib/hammer_cli_foreman/virtual_machine.rb
```diff
@@ -0,0 +1,78 @@
...
+
+      resource :compute_resources, :available_virtual_machines
+      command_name 'list'
```
ofedoren: If the command is `ListCommand` already, there is no need to explicitly specify command's name. Applies for command definitions below as well.

lib/hammer_cli_foreman/virtual_machine.rb
```diff
@@ -0,0 +1,78 @@
...
+    end
+
+    class PowerVmCommand < HammerCLIForeman::Command
+
+      resource :compute_resources, :power_vm
+      command_name 'power'
+
+      option '--compute-resource-id', 'ID', '',
+             :attribute_name => HammerCLI.option_accessor_name('id'), :required => true
+      option "--vm-id", "VM-ID", _("Virtual machine id"), :required => true
```
ofedoren: I personally don't like that we've got the same options defined multiple times in this file.

I'd create a command extension for this purpose (in case we'll need to remove them or somehow adjust in the future).

Moreover, I think we need to add `MultiResourceCommand` definition (which is out of the scope of this PR unless you want to do it, so we could actually test it with those commands), because of redundant/wrong automatically generated options such as `--id` and `--name`. Those options are expecting data for a compute resource... resource, but as a user I'd expect that expected data are for virtual machine (`hammer compute-resource virtual-machine info --id VM-ID`).

test/functional/virtual_machine_test.rb
```diff
@@ -0,0 +1,148 @@
+
+require File.join(File.dirname(__FILE__), 'test_helper')
+
+require 'hammer_cli_foreman/compute_resource'
+
+describe HammerCLIForeman::VirtualMachine do
```
ofedoren: :heart:

lib/hammer_cli_foreman/virtual_machine.rb
yifatmakias: Thanks :) I agree and tried to change my code according to all your comments. I think maybe now it is better and there is no need to add MultiResourceCommand definition. Let me know what do you think about my changes.

lib/hammer_cli_foreman/virtual_machine.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      success_message _("Virtual machine is powering.")
+      failure_message _("Could not power the virtual machine")
+
+      build_options
+    end
+
+    class InfoCommand < HammerCLIForeman::InfoCommand
+      action :show_vm
+      command_name 'info'
```
ofedoren: This is not needed, since `InfoCommand` aready contains that.

lib/hammer_cli_foreman/virtual_machine.rb
```diff
@@ -0,0 +1,57 @@
...
+        field :id, _("Id")
+        field :name, _("Name")
+      end
+
+      build_options
+    end
+
+    class DeleteCommand < HammerCLIForeman::DeleteCommand
+      action :destroy_vm
+      command_name 'delete'
```
ofedoren: This is not needed, since `DeleteCommand` aready contains that.

lib/hammer_cli_foreman/virtual_machine.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      def self.included(base)
+        base.build_options
+
+        base.validate_options do
+          any(:option_compute_resource_id, :option_compute_resource_name).required
+        end
+      end
+
+    end
```
ofedoren: This is not used anywhere. Is that a leftover?

lib/hammer_cli_foreman/virtual_machine.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    class InfoCommand < HammerCLIForeman::InfoCommand
+      action :show_vm
+      command_name 'info'
+
+      output do
+        field :id, _("Id")
+        field :name, _("Name")
+      end
```
ofedoren: There is a lot of information from the server for each virtual machine that we could get, but we show `Id` and `Name` only. Could we show all the information we're getting from the server?

yifatmakias: I actually checked it and it is a bit problematic because in each compute resource the data is named differently. For example, in ovirt the number of CPUs is under: 'cpu' while in vmware it is under: 'cpus'.
Do you maybe have an idea how to overcome this issue?

lib/hammer_cli_foreman/virtual_machine.rb
yifatmakias: You are very right, I added it but did not use it eventually. I will delete this part.

lib/hammer_cli_foreman/virtual_machine.rb
ofedoren: Yes, we've got method called [provider_specific_fields](https://github.com/theforeman/hammer-cli-foreman/blob/d214945f09f88f4270476d162b93fac2c92ff66d/lib/hammer_cli_foreman/compute_resource/base.rb#L1) which allows you specify which fields for which CR should be displayed by [redefining it](https://github.com/theforeman/hammer-cli-foreman/blob/d214945f09f88f4270476d162b93fac2c92ff66d/lib/hammer_cli_foreman/compute_resource/gce.rb#L22).

But this assumes that we've got `'provider'` item in the response.

yifatmakias: Ok, sounds good, I will check if the response contains 'provider' item.

yifatmakias: There isn't a 'provider' item in the response. Is it necessary in order to use the relevant method?

yifatmakias: Not hard at all, one-row addition in the compute-resource API. I will try to add it.

ofedoren:
@yifatmakias, what's the status here?

yifatmakias:
I am working on adding each compute resource its relevant fields for the info command. I already finished adding libvirt, Ovirt, vmware. I am now working on gce, ec2 OpenStack and Rackspace.

yifatmakias:
This PR is now dependent on a foreman PR:
https://github.com/theforeman/foreman/pull/7495

In the PR I mentioned above the provider name was added to the API result in order to add each compute resource its relevant info.

yifatmakias:
We definitely could. Which VM specific fields do you think we should add?

ofedoren:
Unfortunatelly since there is a lot of differences between responses for different providers, I'd suggest two options:

1. Find similar information for each provider (e.g. name, cpus, memory) and add fields accordingly (same field names, but different pathes for example). This will show consistent information across all the providers, but will show less information than the option №2.
2. Add fields for all the information we can gather from the responses. This will show all the information related to a VM, but we will need to add more fields for each provider.

yifatmakias:
I prefer option 2. I will add more fields to each provider.

yifatmakias:
I see your point here, maybe it is a better idea to add more fields under this PR because I already have all the providers setup and could do that quite easily. I will add some fields to each provider.

yifatmakias:
@ofedoren Hi :) I added more fields to each provider, the fields that looked relevant to me. Let me know what you think.

ofedoren:
@yifatmakias, thanks :) LGTM. @shiramax, I see that you've requested some changes. Were your concerns adressed? If so, it's ready to merge.

shiramax:
@ofedoren yes thanks :)

ofedoren:
Thanks, @yifatmakias, @shiramax! Merging.

PR #471, opened 2019-12-31
Fixes #28541 - Fix host creation from image and hostgroup

shiramax:
In the code we are translating the forman image id to the provider image_id and put it under ['host']['compute_attributes']['image_id'], 
the problem is that we relied on compute_resource_id to exist or host_id (in case of update) and we didn't address a commend that send only with hostgroup, 
this PR should fix the issue.

### DIFF
--- a/lib/hammer_cli_foreman/compute_resource/utils.rb
+++ b/lib/hammer_cli_foreman/compute_resource/utils.rb
@@ -8,6 +8,22 @@ module HammerCLIForeman
       )
     end
 
+    def self.get_hostgroup_compute_resource_id(hostgroup_id)
+      hostgroup = HammerCLIForeman.record_to_common_format(
+          HammerCLIForeman.foreman_resource(:hostgroups).call(:show, 'id' => hostgroup_id)
+      )
+      compute_resource_id = hostgroup['compute_resource_id']
+       if !hostgroup['compute_resource_name'].to_s.strip.empty? && compute_resource_id.nil?
+         compute_resource= HammerCLIForeman.record_to_common_format(
+            HammerCLIForeman.foreman_resource(:compute_resources).call(
+                :index, :search => "name = \"#{hostgroup['compute_resource_name']}\""
+            )
+         )
+         compute_resource_id = compute_resource['results'][0]['id'] if compute_resource['results'][0]
+      end
+      compute_resource_id
+    end
+
     def self.get_host_compute_resource_id(host_id)
       HammerCLIForeman.record_to_common_format(
         HammerCLIForeman.foreman_resource(:hosts).call(
--- a/lib/hammer_cli_foreman/hosts/common_update_options.rb
+++ b/lib/hammer_cli_foreman/hosts/common_update_options.rb
@@ -98,9 +98,14 @@ module HammerCLIForeman
           params['host']['compute_attributes']['volumes_attributes'] = nested_attributes(option_volume_list)
           params['host']['interfaces_attributes'] = interfaces_attributes
         end
-
-        if options["option_image_id"]
-          compute_resource_id = params['host']['compute_resource_id'] || ::HammerCLIForeman::ComputeResources.get_host_compute_resource_id(params['id'])
+        if options['option_image_id']
+          if params['host']['compute_resource_id']
+            compute_resource_id = params['host']['compute_resource_id']
+          elsif params['id']
+            compute_resource_id = ::HammerCLIForeman::ComputeResources.get_host_compute_resource_id(params['id'])
+          elsif params['host']['hostgroup_id']
+            compute_resource_id = ::HammerCLIForeman::ComputeResources.get_hostgroup_compute_resource_id(params['host']['hostgroup_id'])
+          end
           raise ArgumentError, "Missing argument for 'compute_resource'" if compute_resource_id.nil?
           image_uuid =  ::HammerCLIForeman::ComputeResources.get_image_uuid(compute_resource_id, options["option_image_id"])
           params['host']['compute_attributes']['image_id'] = image_uuid
--- a/test/functional/host_test.rb
+++ b/test/functional/host_test.rb
@@ -236,6 +236,67 @@ describe "host create" do
     result = run_cmd(cmd + minimal_params_without_hostgroup + params)
     assert_cmd(expected_result, result)
   end
+
+  it "Creates host in case the image is provided from the hostgroup and hostgroup is not nested" do
+    params = ['--name=test',
+              '--provision-method=image',
+              '--image-id=8' ,
+              '--managed=true',
+              '--hostgroup-id=1',
+              '--organization-id=1',
+              '--location-id=1'
+    ]
+
+    api_expects(:hostgroups, :show).with_params(
+      {}
+    ).returns({'compute_resource_name' => "cr", 'compute_resource_id' => 33 })
+
+    api_expects(:images, :show).with_params(
+      {"compute_resource_id" => 33, "id" => 8}
+    ).returns(results: {'uuid' => '111111'})
+
+    api_expects(:hosts, :create).with_params(
+      {"location_id" => 1, "organization_id" => 1, "host" => {"name" => "test", "location_id" => 1, "organization_id" => 1, "puppetclass_ids" => [], "hostgroup_id" => 1, "image_id" => 8, "provision_method" => "image", "managed" => true, "compute_attributes" => {"volumes_attributes" => {}, "image_id" => nil}, "build" => true, "enabled" => true, "overwrite" => true, "interfaces_attributes" => []}}
+    ).returns(results: {'uuid' => '111111'})
+
+    expected_result = success_result("Host created.\n")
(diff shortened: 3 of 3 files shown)


kgaikwad:
@ofedoren, any updates on this?

lib/hammer_cli_foreman/hosts/common_update_options.rb
```diff
@@ -100,7 +100,13 @@ def request_params
...
 
         if options["option_image_id"]
-          compute_resource_id = params['host']['compute_resource_id'] || ::HammerCLIForeman::ComputeResources.get_host_compute_resource_id(params['id'])
+          if params['host']['compute_resource_id']
+            compute_resource_id = params['host']['compute_resource_id']
+          elsif  params['id']
+            compute_resource_id = ::HammerCLIForeman::ComputeResources.get_host_compute_resource_id(params['id'])
+          elsif params['host']['hostgroup_id']
+            compute_resource_id = ::HammerCLIForeman::ComputeResources.get_hostgroup_compute_resource_id(params['host']['hostgroup_id'])
+          end
```
rahulbajaj0509: Nit: Sorry for adding this nit, it is a personal choice and feel free to ignore :)
When we have multiple if/else/ifelse, we can use the `case` statement. Something like:

```
if options['option_image_id']
  compute_resource_id = case [params['host']['compute_resource_id'], ::HammerCLIForeman::ComputeResources.get_host_compute_resource_id(params['id'])]
                          when params['host']['compute_resource_id'] then params['host']['compute_resource_id']
                          when params['id'] then ::HammerCLIForeman::ComputeResources.get_host_compute_resource_id(params['id'])
                          when params['host']['hostgroup_id'] then ::HammerCLIForeman::ComputeResources.get_hostgroup_compute_resource_id(params['host']['hostgroup_id'])
                        end
  raise ArgumentError, "Missing argument for 'compute_resource'" if compute_resource_id.nil?
  image_uuid =  ::HammerCLIForeman::ComputeResources.get_image_uuid(compute_resource_id, options["option_image_id"])
  params['host']['compute_attributes']['image_id'] = image_uuid
end

I haven't tested this, but it looks like it. @shiramax @kgaikwad let me know your thoughts on this!

Comment (shiramax) [reply lib/hammer_cli_foreman/hosts/common_update_options.rb]: thanks @rahulbajaj0509 , Done :)

File: lib/hammer_cli_foreman/hosts/common_update_options.rb [hunk 2]
```diff
@@ -98,9 +98,12 @@ def request_params
           params['host']['compute_attributes']['volumes_attributes'] = nested_attributes(option_volume_list)
           params['host']['interfaces_attributes'] = interfaces_attributes
         end
-
-        if options["option_image_id"]
-          compute_resource_id = params['host']['compute_resource_id'] || ::HammerCLIForeman::ComputeResources.get_host_compute_resource_id(params['id'])
+        if options['option_image_id']
+          compute_resource_id = case
```
kgaikwad: @shiramax, 

There is no variable or regex inside case statement then against which value `when` statements will do matching? 
You need to mention something inside case statement otherwise it will take first value from first when statement.
Please correct me if I am wrong.

lib/hammer_cli_foreman/hosts/common_update_options.rb
kgaikwad: @rahulbajaj0509, @shiramax 

In my opinion, +1 to if/else/ifelse loop that was there previously.
It was more readable.

lib/hammer_cli_foreman/hosts/common_update_options.rb
kgaikwad: For readability, it would be good to have simple `if/else/ifelse` as I mentioned.

shiramax: @kgaikwad 
from what I understand it possible to create a switch case statement without the case like follows: 

case
when a < 3
  "Smaller than 3"
end

and I also tested it, and it works fine to me, can you confirm it works for you as well ?

lib/hammer_cli_foreman/hosts/common_update_options.rb
```diff
@@ -98,9 +98,14 @@ def request_params
           params['host']['compute_attributes']['volumes_attributes'] = nested_attributes(option_volume_list)
           params['host']['interfaces_attributes'] = interfaces_attributes
         end
-
-        if options["option_image_id"]
-          compute_resource_id = params['host']['compute_resource_id'] || ::HammerCLIForeman::ComputeResources.get_host_compute_resource_id(params['id'])
+        if options['option_image_id']
+          if params['host']['compute_resource_id']
+            compute_resource_id = params['host']['compute_resource_id']
+          elsif  params['id']
```
ofedoren: Nit: extra space after `elsif`.

lib/hammer_cli_foreman/compute_resource/utils.rb
```diff
@@ -8,6 +8,14 @@ def self.get_image_uuid(compute_resource_id, image_id)
       )
     end
 
+    def self.get_hostgroup_compute_resource_id(hostgroup_id)
+      HammerCLIForeman.record_to_common_format(
+          HammerCLIForeman.foreman_resource(:hostgroups).call(
+              :show, 'id' => hostgroup_id
+          )['compute_resource_id']
```
ofedoren: After some testing I've noticed, that if we use host group that is nested then the server response contains only the name of a compute resource (`"compute_resource_id" => nil`), so this will cause error `Missing argument for 'compute_resource'`.

What do you think will be better:
  1. Support resource id resolving in case we have the resource name only?
  2. Fix this on API side.

shiramax: @ofedoren good catch! thanks @ofedoren.
I'm working on a fix for that in foreman-core. I will update this one as soon as I finish with the PR .
 thanks

shiramax:
@ofedoren thanks !
I fixed the issue in hammer, because it's faster but I've opened another for foreman, and when he will ve merged I will change it here as well. 
I've also added some test :) 
It would be great if you can review it today so we would be able to release a new hammer version today. 
thanks :)

lib/hammer_cli_foreman/compute_resource/utils.rb
```diff
@@ -8,6 +8,21 @@ def self.get_image_uuid(compute_resource_id, image_id)
       )
     end
 
+    def self.get_hostgroup_compute_resource_id(hostgroup_id)
+      hostgroup = HammerCLIForeman.record_to_common_format(
+          HammerCLIForeman.foreman_resource(:hostgroups).call(:show, 'id' => hostgroup_id)
+      )
+      compute_resource_id = hostgroup['compute_resource_id']
+      unless compute_resource_id
```
tbrisker: should this be something like?
```suggestion
      if hostgroup['compute_resource_name'].present? && compute_resource_id.nil?
```
so we don't do the extra query if there is no compute resource?

lib/hammer_cli_foreman/compute_resource/utils.rb
```diff
@@ -8,6 +8,21 @@ def self.get_image_uuid(compute_resource_id, image_id)
...
+
+        compute_resources = HammerCLIForeman.record_to_common_format(
+            HammerCLIForeman.foreman_resource(:compute_resources).call(:index)
```
tbrisker: should we add a search for the name instead of getting the entire list (which might be paginated) and looking for the id in it?

lib/hammer_cli_foreman/compute_resource/utils.rb
kgaikwad: In case no ActiveSupport extension, then
```suggestion
      if hostgroup['compute_resource_name'].to_s.strip.empty? && compute_resource_id.nil?
```

shiramax:
@tbrisker @kgaikwad done, thanks!

lib/hammer_cli_foreman/compute_resource/utils.rb
```diff
@@ -8,6 +8,21 @@ def self.get_image_uuid(compute_resource_id, image_id)
...
+      hostgroup = HammerCLIForeman.record_to_common_format(
+          HammerCLIForeman.foreman_resource(:hostgroups).call(:show, 'id' => hostgroup_id)
+      )
+      compute_resource_id = hostgroup['compute_resource_id']
+       if !hostgroup['compute_resource_name'].to_s.strip.empty? && compute_resource_id.nil?
+         compute_resource_id = HammerCLIForeman.record_to_common_format(
+            HammerCLIForeman.foreman_resource(:compute_resources).call(
+                :index, :search => "name = \"#{hostgroup['compute_resource_name']}\""
+            )
+         )['results'][0]['id']
```
tbrisker: what if no matching resource are found?  `['results'][0]` will be `nil` i think?

shiramax: @tbrisker fixed :)

PR #473, opened 2019-12-31
Fixes #28599 - drop smart variable handling

yifatmakias:
This PR is related to this hammer pr:
https://github.com/theforeman/hammer-cli-foreman/pull/469

it should be merged before merging the second pr because the foreman-API file is updated in this pr and the second pr is related to the API changes.


orrabin:
The new `foreman_api.json` should be for 2.0, not replacing 1.24 since this does not reflect 1.24.

shiramax:
thanks @yifatmakias , good work!

PR #476, opened 2020-01-04
Unpin mocha version & fix deprecation warning

floehopper:
The regression introduced in mocha v1.10.0 which led to its version being pinned to v1.9.0 was fixed in mocha v1.11.2 which is currently the latest version, so it is now safe to unpin the version.

Requiring 'mocha/setup' was deprecated in v1.10.0, so I've also fixed that with the change to `test/test_helper.rb`.

Fixes [Redmine issue 28420][1].

[1]: https://projects.theforeman.org/issues/28420

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,7 @@ group :test do
   gem 'minitest', '4.7.4'
   gem 'minitest-spec-context'
   gem 'simplecov'
-  gem 'mocha', '<=1.9.0'
+  gem 'mocha'
   gem 'ci_reporter', '>= 1.6.3', "< 2.0.0", :require => false
 end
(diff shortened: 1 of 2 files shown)


mmoll:
ok to test

mmoll:
@floehopper Thanks, could you please prepend "Fixes #28420" to the commit message, then https://projects.theforeman.org/issues/28420 will be correctly handled.

floehopper:
@mmoll:

> @floehopper Thanks, could you please prepend "Fixes #28420" to the commit message, then https://projects.theforeman.org/issues/28420 will be correctly handled.

Done & force-pushed.

PR #478, opened 2020-01-06
Fixes #28646 - fix unauthorized user error message

rahulbajaj0509:
Trying to login with incorrect credentials when a user is already logged in, the error message shows something like: invalid username and password. Since we have multiple auth source types present now, we can change the message to something like: invalid credentials.


xprazak2:
Makes sense, test failures related.

PR #480, opened 2020-01-09
Fixes #28704 - add auth source external command

rahulbajaj0509:


lib/hammer_cli_foreman/auth_source_external.rb
```diff
@@ -0,0 +1,30 @@
...
+    resource :auth_source_externals
+    command_name 'external'
+    desc _('Manage external auth sources')
+
+    class ListCommand < HammerCLIForeman::ListCommand
+      output do
+        field :id, _('Id')
+        field :name, _('Name')
+        field :location_id, _('Location')
+        field :organization_id, _('Organization')
```
rahulbajaj0509: @ofedoren while I am testing the PR, I am unable to get the location and organization values in the list command. Can you suggest a way in which users can see the same?

lib/hammer_cli_foreman/auth_source.rb
```diff
@@ -18,16 +19,6 @@ class ListCommand < HammerCLIForeman::ListCommand
...
-    class UpdateCommand < HammerCLIForeman::UpdateCommand
-      desc _('Update organization and location for Auth Source')
-      resource :auth_source_externals
-  
-      success_message _('Taxonomy updated.')
-      failure_message _('Taxonomy not changed. Please check if appropriate auth source exists')
-
-      build_options
-    end
-
```
rahulbajaj0509: Removing this and created a new subcommand altogether, similar to the ldap subcommand. I hope this makes sense!

lib/hammer_cli_foreman/auth_source_external.rb
rahulbajaj0509: Or I would require the info command to show the same?

lib/hammer_cli_foreman/auth_source.rb
ofedoren: DRYing out always makes sense :) But does it make sense to leave this command for backwards compatibility also?

If so, we could change the whole command definition to:
```ruby
    subcommand 'external', ...
    subcommand 'update', HammerCLIForeman::AuthSourceExternal::UpdateCommand.desc, HammerCLIForeman::AuthSourceExternal::UpdateCommand
    # ...
```

OR we leave this command as it is, but add the pretty the same line in the `HammerCLIForeman::AuthSourceExternal` command class definition.

lib/hammer_cli_foreman/auth_source_external.rb
ofedoren: Yes, I've noticed that too. This is because the response from the server contains `organizations/locations` collections instead of single org/loc. So the best way here is to change those two fields to `HammerCLIForeman::References.taxonomies(self)`. BUT this will not work either for `table` output adapter because taxonomies fields will be cleared from output fields because of hammer core design (since collections contain more than one piece of data for a field, we don't show them in tables because output can be broken).

This is what I'm suggesting:
  1. Since external auth source can have multiple org/loc, change two fields as I mentioned, so the taxonomies are shown at least if `--output=base` is used.
  2. Change the default adapter to be `base`, so the taxonomies are shown always (but this will not be consistent with other list commands' output).
  3. OR instead of 1 and 2 implement `List` command without taxonomies and implement `Info` command with taxonomies.

I'd prefer the 3'd way, because it will be consistent with other commands and also we implement `Info` command for the endpoint we have.

rahulbajaj0509: @ofedoren: myself and @shiramax were also talking about it and figured out the same thing as you suggested. Looks like we are all on the same page on this! 

Will update the PR with all the changes and add test cases to it :) thank you both.

lib/hammer_cli_foreman/auth_source.rb
rahulbajaj0509: I added this command like 3 days ago and realized it was not the correct way, so i thought it would be ohkay not caring too much about it. Would that be ohkay for us or we want to do something like adding the subcommand. Let me know if you think otherwise :)

lib/hammer_cli_foreman/auth_source_external.rb
```diff
@@ -0,0 +1,36 @@
...
+      end
+
+      build_options
+    end
+
+    class UpdateCommand < HammerCLIForeman::UpdateCommand
+      desc _('Update organization and location for Auth Source')
+      resource :auth_source_externals
+
+      success_message _('Taxonomy updated.')
```
rahulbajaj0509: Users can also change the name of the auth source here, therefore, changing the message to `Successfully updated the external auth source`. Sounds better?

rahulbajaj0509:
I would like to add functional tests for LDAP and External users but I want to have it in another PR. Can we merge this one without the functional ones if you are ok? @ofedoren

ofedoren:
@rahulbajaj0509, sure. Otherwise it's ready to be merged anyway :)

Merging.

PR #484, opened 2020-01-13
Fixes #19999 - support associating multiple provisioning templates

yifatmakias:
… once

### DIFF
--- a/lib/hammer_cli_foreman/associating_commands.rb
+++ b/lib/hammer_cli_foreman/associating_commands.rb
@@ -234,10 +234,37 @@ module HammerCLIForeman
 
       class AddProvisioningTemplateCommand < HammerCLIForeman::AddAssociatedCommand
         associated_resource :provisioning_templates
-        desc _("Associate a provisioning template")
+        desc _("Associate provisioning templates")
 
-        success_message _("The provisioning template has been associated.")
-        failure_message _("Could not associate the provisioning template")
+        option "--provisioning-template-ids", "PROVISIONING_TEMPLATE_IDS", _("List of provisioning template ids"),
+               format: HammerCLI::Options::Normalizers::List.new
+        option "--provisioning-templates", "PROVISIONING_TEMPLATE_NAMES", _("List of provisioning template names"),
+               format: HammerCLI::Options::Normalizers::List.new, attribute_name: :option_provisioning_template_names
+        option "--provisioning-template-search", "PROVISIONING_TEMPLATE_SEARCH", _("Provisioning template name regex to search, all matching templates will be associated")
+
+        def request_params
+          params = super
+
+          if options['option_provisioning_template_search']
+            templates = HammerCLIForeman.collection_to_common_format(
+              associated_resource.call(
+                :index,
+                :search => "name ~ \"#{options['option_provisioning_template_search']}\"")
+            )
+            raise ResolverError.new(_("%s not found.") % associated_resource.singular_name, associated_resource) if templates.empty?
+
+            templates.each do |template|
+              template_id = template['id']
+              params['operatingsystem']['provisioning_template_ids'] << template_id.to_s
+            end
+          end
+          params['operatingsystem']['provisioning_template_ids'] = params['operatingsystem']['provisioning_template_ids'].uniq
+          params
+        end
+
+        success_message _("The provisioning templates were associated.")
+        failure_message _("Could not associate the provisioning templates")
+
       end
 
       class RemoveProvisioningTemplateCommand < HammerCLIForeman::RemoveAssociatedCommand
--- a/test/functional/associating_commands_test.rb
+++ b/test/functional/associating_commands_test.rb
@@ -1,0 +1,106 @@
+require File.join(File.dirname(__FILE__), 'test_helper')
+
+require 'hammer_cli_foreman/operating_system'
+
+describe HammerCLIForeman::OperatingSystem do
+  context "AddProvisioningTemplateCommand" do
+    before do
+      @cmd = ["os", "add-provisioning-template"]
+      @os_before_update = {
+        id: 1,
+        provisioning_templates: []
+      }
+      @os_after_update1 = {
+        id: 1,
+        provisioning_templates: [{ id: 1, name: 'test template' }]
+      }
+      @os_after_update2 = {
+        id: 1,
+        provisioning_templates: [{ id: 1, name: 'test template 1' },
+                                 { id: 2, name: 'test template 2' }]
+      }
+      @search_result = [{ 'id' => 1, 'name' => 'test template 1'}, { 'id' => 2, 'name' => 'test template 2'}]
+      @provisioning_templates = [{ 'id' => 1, 'name' => 'test template 1' }, { 'id' => 2, 'name' => 'test template 2' }]
+    end
+
+    it "should print error on missing --id" do
+      expected_result = "Could not associate the provisioning templates:\n  Missing arguments for '--id'\n"
+
+      api_expects_no_call
+      result = run_cmd(@cmd)
+      assert_match(expected_result, result.err)
+    end
+
+    it "should associate the given provisioning template-by id" do
+      params = ['--id=1', '--provisioning-template-id=1']
+
+      api_expects(:operatingsystems, :show).with_params({ "id": "1" }).returns(@os_before_update)
+
+      api_expects(:operatingsystems, :update) do |par|
+        par['id'] == '1' &&
+          par['operatingsystem'] == { 'provisioning_template_ids' => ['1'] }
+      end.returns(@os_after_update1)
+
+      result = run_cmd(@cmd + params)
+      assert_cmd(
+        success_result("The provisioning templates were associated.\n"),
+        result
+      )
+    end
+
+    it "should associate all given provisioning templates - by ids" do
+      params = ['--id=1', '--provisioning-template-ids=1,2']
+
+      api_expects(:operatingsystems, :show).with_params({ "id": "1" }).returns(@os_before_update)
+
+      api_expects(:operatingsystems, :update) do |par|
+        par['id'] == '1' &&
+          par['operatingsystem'] == { 'provisioning_template_ids' => ['1', '2'] }
+      end.returns(@os_after_update2)
+
+      result = run_cmd(@cmd + params)
+      assert_cmd(
+        success_result("The provisioning templates were associated.\n"),
+        result
+      )
+    end
+
+    it "should associate all given provisioning templates - by names" do
+      params = ['--id=1', '--provisioning-templates=test template 1,test template 2']
+
+      api_expects(:operatingsystems, :show).with_params({ "id": "1" }).returns(@os_before_update)
+      api_expects(:provisioning_templates, :index).with_params(
+        :search => "name = \"test template 1\" or name = \"test template 2\"",
+        :per_page => 1000,
+        :page => 1).returns(@search_result)
+      api_expects(:operatingsystems, :update) do |par|
+        par['id'] == '1' &&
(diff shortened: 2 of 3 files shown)


lib/hammer_cli_foreman/associating_commands.rb
```diff
@@ -236,8 +236,42 @@ class AddProvisioningTemplateCommand < HammerCLIForeman::AddAssociatedCommand
         associated_resource :provisioning_templates
         desc _("Associate a provisioning template")
```
xprazak2: Should we change description to 'Associate provisioning templates'?

lib/hammer_cli_foreman/associating_commands.rb
```diff
@@ -236,8 +236,42 @@ class AddProvisioningTemplateCommand < HammerCLIForeman::AddAssociatedCommand
...
 
+        option "--provisioning-template-ids", "PROVISIONING_TEMPLATE_IDS", _("List of provisioning template ids"),
+               format: HammerCLI::Options::Normalizers::List.new
+        option "--provisioning-templates", "PROVISIONING_TEMPLATE_NAMES", _("List of provisioning template names"),
+               format: HammerCLI::Options::Normalizers::List.new
+        option "--provisioning-template-search", "PROVISIONING_TEMPLATE_SEARCH", _("Provisioning template name regex to search, all matching templates will be associate")
```
xprazak2: nit: all matching templates will be associated

lib/hammer_cli_foreman/associating_commands.rb
```diff
@@ -236,8 +236,42 @@ class AddProvisioningTemplateCommand < HammerCLIForeman::AddAssociatedCommand
...
+
+        def request_params
+          params = super
+          template_ids = params['operatingsystem']['provisioning_template_ids']
+          if options['option_provisioning_template_ids']
+            template_ids |= options['option_provisioning_template_ids']
+          end
+          if options['option_provisioning_templates']
+            options['option_provisioning_templates']&.each do |provisioning_template_name|
```
xprazak2: You are also checking for nil on the previous line, I think it is ok to omit one of the checks.

lib/hammer_cli_foreman/associating_commands.rb
```diff
@@ -236,8 +236,42 @@ class AddProvisioningTemplateCommand < HammerCLIForeman::AddAssociatedCommand
...
+              template_id = HammerCLIForeman.collection_to_common_format(
+                associated_resource.action(:index).call(
+                  search: "name = \"#{provisioning_template_name}\"", per_page: 1000, page: 1)).first['id']
```
xprazak2: Raises `NoMethodError (undefined method `[]' for nil:NilClass)` when no template with `provisioning_template_name` is found.

lib/hammer_cli_foreman/associating_commands.rb
```diff
@@ -236,8 +236,42 @@ class AddProvisioningTemplateCommand < HammerCLIForeman::AddAssociatedCommand
...
+              template_ids << template_id.to_s
+            end
+          end
+          if options['option_provisioning_template_search']
+            provisioning_templates = HammerCLIForeman.collection_to_common_format(
+              associated_resource.action(:index).call(per_page: 1000, page: 1)
+            )
+            matching_provisioning_templates = provisioning_templates.select do |ct|
+              ct['name'].match(options['option_provisioning_template_search'])
```
xprazak2: Would it make sense to call api search in a similar way like you have for `--provisioning-templates`? The [~](https://github.com/wvanbergen/scoped_search/wiki/Query-language#comparison-operators) operator should do the job, but it is case-insensitive. If we want case-sensitive, I think it is ok to keep as is.

lib/hammer_cli_foreman/associating_commands.rb
```diff
@@ -236,8 +236,42 @@ class AddProvisioningTemplateCommand < HammerCLIForeman::AddAssociatedCommand
...
+            end
+            template_ids |= matching_provisioning_templates.map { |ct| ct['id'].to_s }
+          end
+          params['operatingsystem']['provisioning_template_ids'] = template_ids.reject(&:empty?)
```
xprazak2: The ids from user input are ints and do not have `empty?` defined:

empty-arr

lib/hammer_cli_foreman/associating_commands.rb
```diff
@@ -236,8 +236,42 @@ class AddProvisioningTemplateCommand < HammerCLIForeman::AddAssociatedCommand
...
+          params
+        end
+
         success_message _("The provisioning template has been associated.")
         failure_message _("Could not associate the provisioning template")
```
xprazak2: Should we pluralize the messages here as well?

xprazak2:
We probably do not want to pluralize the subcommand and keep options for a single template for backward compatibility, right?

lib/hammer_cli_foreman/associating_commands.rb
ofedoren: Also, we don't use `&.` since this comes with Ruby 2.3+, but hammer could be used on systems with lower version of Ruby.

Honestly, I'd review the minimal supported version of hammer, what do you think about that, @shiramax?

yifatmakias:
I tried to fix it according to all the reviews you gave me. I hope it is OK now.
I did not understand your last comment? Do you think we should remove the options for a single template?

lib/hammer_cli_foreman/associating_commands.rb
yifatmakias: I fixed it, it should not happen again.

lib/hammer_cli_foreman/associating_commands.rb
```diff
@@ -234,10 +234,48 @@ module ProvisioningTemplate
 
       class AddProvisioningTemplateCommand < HammerCLIForeman::AddAssociatedCommand
         associated_resource :provisioning_templates
-        desc _("Associate a provisioning template")
+        desc _("Associate provisioning templates")
+
+        option "--provisioning-template-ids", "PROVISIONING_TEMPLATE_IDS", _("List of provisioning template ids"),
+               format: HammerCLI::Options::Normalizers::List.new
+        option "--provisioning-templates", "PROVISIONING_TEMPLATE_NAMES", _("List of provisioning template names"),
+               format: HammerCLI::Options::Normalizers::List.new
```
ofedoren: I'd suggest to add here `attribute_name: :option_provisioning_template_names` as well, so we can slightly modify [AddAssociatedCommand command](https://github.com/theforeman/hammer-cli-foreman/blob/f163dd9fc9a232f711621f81a16752624edd5de7/lib/hammer_cli_foreman/commands.rb#L668) like so:

```ruby
    def get_new_ids
      ids = get_current_ids.map(&:to_s)
      required_ids = get_associated_identifiers.map(&:to_s)
      required_ids << get_associated_identifier.to_s

      ids += required_ids
      ids.uniq
    end
```

Sure, we'll need to [add](https://github.com/theforeman/hammer-cli-foreman/blob/f163dd9fc9a232f711621f81a16752624edd5de7/lib/hammer_cli_foreman/commands.rb#L617) a new method for retrieving ids:

```ruby
    def get_associated_identifiers
      get_resource_ids(associated_resource, :scoped => true)
    end
```

lib/hammer_cli_foreman/associating_commands.rb
```diff
@@ -234,10 +234,48 @@ module ProvisioningTemplate
...
+            options['option_provisioning_templates'].each do |provisioning_template_name|
+              template = HammerCLIForeman.collection_to_common_format(
+                associated_resource.action(:index).call(
+                  search: "name = \"#{provisioning_template_name}\"", per_page: 1000, page: 1))
+              raise ResolverError.new(_("%s not found.") % associated_resource.singular_name, associated_resource) if template.empty?
+
+              template_id = template.first['id']
+              template_ids << template_id.to_s
+            end
+          end
```
ofedoren: We can avoid this repetitive (I'm quite sure we might need the same code in the future). Please, see my previous comment how.

lib/hammer_cli_foreman/associating_commands.rb
```diff
@@ -234,10 +234,48 @@ module ProvisioningTemplate
...
+            templates = HammerCLIForeman.collection_to_common_format(
+              associated_resource.action(:index).call(
+                search: "name ~ \"#{options['option_provisioning_template_search']}\"", per_page: 1000, page: 1))
+            raise ResolverError.new(_("%s not found.") % associated_resource.singular_name, associated_resource) if templates.empty?
+
+            templates.each do |template|
+              template_id = template['id']
+              template_ids << template_id.to_s
+            end
+          end
```
ofedoren: I'd like to avoid touching `request_params` without a real need.

I'm not aware of any automated way to accept such options, but I think we can use this opportunity to implement such in one of (probably a new one) `option_sources`.

In case you don't want to do it (I think this can be out of the scope of this PR), please adjust the searching at least (remove pagination, so we can search in al thel data, not jsut the first one page, see https://github.com/theforeman/hammer-cli-foreman/blob/f163dd9fc9a232f711621f81a16752624edd5de7/lib/hammer_cli_foreman/compute_resource/utils.rb#L19).

lib/hammer_cli_foreman/associating_commands.rb
ofedoren: After suggested changes I don't think we will face the same issue, so we might not need the `.reject(&:empty?)`.

test/functional/associating_commands_test.rb
```diff
@@ -0,0 +1,109 @@
+require File.join(File.dirname(__FILE__), 'test_helper')
+
+require 'hammer_cli_foreman/operating_system'
+
+describe HammerCLIForeman::OperatingSystem do
+  context "AddProvisioningTemplateCommand" do
```
ofedoren: :heart:

lib/hammer_cli_foreman/associating_commands.rb
```diff
@@ -234,10 +234,36 @@ module ProvisioningTemplate
...
+                :search => "name ~ \"#{options['option_provisioning_template_search']}\"")
+            )
+            raise ResolverError.new(_("%s not found.") % associated_resource.singular_name, associated_resource) if templates.empty?
+
+            templates.each do |template|
+              template_id = template['id']
+              params['operatingsystem']['provisioning_template_ids'] << template_id.to_s
+            end
+          end
+          params
```
ofedoren: I've noticed that if we use e.g. `...search` option that returns an id of a tempalte that was already associated then we've got the same id twice. Should we call `.uniq` on `params['operatingsystem']['provisioning_template-ids']` before return?

yifatmakias: Yes, I think it is a good idea. I will add it.

lib/hammer_cli_foreman/associating_commands.rb
```diff
@@ -234,10 +234,37 @@ module ProvisioningTemplate
...
+              associated_resource.call(
+                :index,
+                :search => "name ~ \"#{options['option_provisioning_template_search']}\"")
+            )
+            raise ResolverError.new(_("%s not found.") % associated_resource.singular_name, associated_resource) if templates.empty?
+
+            templates.each do |template|
+              template_id = template['id']
+              params['operatingsystem']['provisioning_template_ids'] << template_id.to_s
+              params['operatingsystem']['provisioning_template_ids'].uniq
```
ofedoren: This line just calls `.uniq` in ids, but doesn't save the changes. And also this is called on every iteration, which is not quite good.

I'd suggest to add safe `params['operatingsystem']['provisioning_template_ids'] = params['operatingsystem']['provisioning_template_ids'].uniq` right before `params` returns.

PR #485, opened 2020-01-14
Fixes #28747 - remove deprecated permissions api parameters

yifatmakias:


xprazak2:
You are removing the output fields for the `filter available-permissions` command, so now the only remaining field is 'id':

ids-only

Using `name` and `resource_type` as request parameters was [deprecated](https://github.com/theforeman/foreman/pull/2595) and [removed](https://github.com/theforeman/foreman/commit/dbbd5c9c1809b2baae326837aa60a8269ae426b5#diff-f1c8a83462797db341ffd1b83258c4e5) (search should be used instead), but no changes were made for the response - I can still see resource_type and name for permission:

available-permissions

I do not think that changes on foreman side impact hammer because hammer is not using the removed attributes as request params and still gets the same data back. Are we sure we want to remove the output fields?

yifatmakias:
I agree that the output fields should not be removed.

xprazak2:
Thanks!

PR #486, opened 2020-01-15
Fixes #28765 - Fix help for filter create command

ofedoren:


xprazak2:
Fixes the issue for me. I do not want to be a rubocop, but I'd rather avoid reusing the same name for nested block parameters to avoid any confusion. Could one of the two `h`s be renamed please?

ofedoren:
@xprazak2, yeah, that does make sense, since the usage of single `h` could be misleading and confusing. I've changed that to be more explicit. Also, I'm afraid we need someday review all those help extensions and do changes there as well...

xprazak2:
Thanks!

PR #488, opened 2020-01-20
Fixes #12471 - add command for managing trends

yifatmakias:


lib/hammer_cli_foreman/trend.rb
```diff
@@ -0,0 +1,45 @@
+module HammerCLIForeman
+
+  class Trend < HammerCLIForeman::Command
+
+    resource :trends
+
+    class ListCommand < HammerCLIForeman::ListCommand
+      output do
+        field :id, _("Id")
+        field :label, _("Name")
```
ofedoren: I think this should be `:name` field as it shows the name of a trend. The same for `InfoCommand`.

lib/hammer_cli_foreman/trend.rb
```diff
@@ -0,0 +1,45 @@
...
+      end
+
+      build_options
+    end
+
+    class CreateCommand < HammerCLIForeman::CreateCommand
+      success_message _("Trend %<trendable_type>s created.")
```
ofedoren: If the name is optional, I think it makes sense to use the type, but if we use the type, shouldn't we change the message to `Trend for %<trendable_type> created.`? The same for `DeleteCommand`.

test/unit/trend_test.rb
```diff
@@ -0,0 +1,30 @@
+require File.join(File.dirname(__FILE__), 'test_helper')
+require File.join(File.dirname(__FILE__), 'apipie_resource_mock')
+
+require 'hammer_cli_foreman/trend'
+
+describe HammerCLIForeman::Trend do
```
ofedoren: Although :heart: for adding the tests, but If we add new commands, we create the functional tests instead of unit. This is because we rely on apidocs we're generating commands from. If the docs were changed somehow in the future we can easily see that something is broken. It might require updating `test/data` as well.

lib/hammer_cli_foreman/trend.rb
yifatmakias: Do you think it should be :name although the name field is optional?

test/unit/trend_test.rb
yifatmakias: Got it, I will move it to functional tests.

lib/hammer_cli_foreman/trend.rb
```diff
@@ -0,0 +1,45 @@
+module HammerCLIForeman
+
+  class Trend < HammerCLIForeman::Command
+
+    resource :trends
+
+    class ListCommand < HammerCLIForeman::ListCommand
+      output do
```
ofedoren: Since the name is optional for trend creation, I'd suggest to add here :label field as well.

lib/hammer_cli_foreman/trend.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    class DeleteCommand < HammerCLIForeman::DeleteCommand
+      success_message _("Trend for %<trendable_type>s deleted.")
+      failure_message _("Could not delete the trend")
+
+      build_options
+    end
+
+    class InfoCommand < HammerCLIForeman::InfoCommand
+      output do
```
ofedoren: dtto `ListCommand`.

PR #489, opened 2020-01-27
Refs #28440 - Help contains squeezed options

ofedoren:
Requires https://github.com/theforeman/hammer-cli/pull/323.


xprazak2:
Mostly does what it is supposed to, when looking at the output of `hammer host create --help`, some of the options are not squeezed and squeezed options miss description.

ofedoren:
> some of the options are not squeezed and squeezed options miss description.

Yes, that's the problem. The changes for not automatically generated options must be applied by hands (custom options that come from command extensions/plugins).

The descriptions are missing only in `-h` help, in `hammer full-help` they are present. And I'm not quite sure if we want it that way (I do :smile: though). Descriptions in `-h` output are really repetitive, like 
```sh
--location LOCATION        Location name
```
So I'd prefer to not show them, especially if we have e.g. `--location[-id]` then we need to update the description not to be `Location name`, but `Location name or id`, which is the same as the option name itself...

ofedoren:
@shiramax, thanks! Updated.

1. I've got architecture squeezed in `host create -h` already... As well as config groups.
2. I've add a tweak, so now we grab API descriptions more. So, there is no need in stuff like that https://github.com/theforeman/hammer-cli-foreman/blob/6ad05e8a2c8c69d412fc222973c1ab540aa8a5a1/lib/hammer_cli_foreman/id_resolver.rb#L56

> would it make sense to add all the resources name, for example:
-architecture[-id | -name]
so only what in the brackets will be valid?

I guess it will break automatization for users if we suddenly change all options like `--resource` to `--resource-name`. Or we would need to hold for some releases the old option.

> How the user would know that --config-group is not an option?

For old users nothing will change in terms of using options. But I agree, for new users the new help might be misleading, but after the first try I hope it will be clear (if a user uses e.g. `-- config-group` instead of `--config-groups` an error will be raised).

shiramax:
@ofedoren 

> For old users nothing will change in terms of using options. But I agree, for new users the new help might be misleading, but after the first try I hope it will be clear (if a user uses e.g. `-- config-group` instead of `--config-groups` an error will be raised).

ok, I agree, and PR overall looks very good!
just there are  some misleading descriptions, for eample: 

host create --help: 

--config-group[s|-ids]                        IDs of associated config groups
is there an option to change the description to "IDs/Names of associated config groups" or something more general like "identifiers of associated config groups"

ofedoren:
@shiramax, updated. I've tried to do some automationed general descriptions for such cases. Requires to update local checkout of PR for hammer-cli as well.

shiramax:
@ofedoren, it looks great!!!
there are a couple of options that are not in the new format of "name/id" :
hammer host create --help

  --puppet-ca-proxy[-id]                        Puppet CA proxy ID
 --puppet-class[es|-ids]                       Comma separated list of values. Values containing comma should be quoted or escaped with backslash.
                                               JSON is acceptable and preferred way for complex parameters
 --puppet-proxy[-id]                           Puppet proxy ID

ofedoren:
@shiramax, thanks! Updated in hammer core as well.

shiramax:
[test hammer]

shiramax:
thanks @ofedoren , can you rebase, so the tests will run again ?

ofedoren:
@shiramax, done :)

shiramax:
thanks @ofedoren !

tbrisker:
Is it possible the nightly bats tests are failing due to this change? it seems to fail at https://github.com/theforeman/forklift/blob/db0b263/bats/fb-test-puppet.bats#L81-L85 with:
```
# (in test file fb-test-puppet.bats, line 84)
#   `hammer host update $pc_ids $id --name $(hostname -f)' failed with status 64
# Could not update the host:
#   Error: Too many arguments.
```
which caused me to think maybe the change in help text caused the script to generate the wrong command?

shiramax:
@tbrisker yes, the help was changed to puppet-class[es|-ids]. 
I saw you already fix it, thanks!

PR #496, opened 2020-02-11
Fixes #28984 - Adding location inherited parameters

shiramax:
Depend on: 
https://github.com/theforeman/foreman/pull/7434

### DIFF
--- a/lib/hammer_cli_foreman/location.rb
+++ b/lib/hammer_cli_foreman/location.rb
@@ -28,14 +28,23 @@ module HammerCLIForeman
       output ListCommand.output_definition do
         field nil, _("Parent"), Fields::SingleReference, :key => :parent
         HammerCLIForeman::References.users(self)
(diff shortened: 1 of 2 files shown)


xprazak2:
Works well, lets wait until the changes in foreman are merged.

shiramax:
@ofedoren thanks!
I'm not sure about this fix, I think it's a bit less readable, but it does remove a lot of repetitive code, 
If we will decide to choose your suggested fix, what do you think about all the parameters that are not Inherited? Do you think we should address them in the class as well?

ofedoren:
@shiramax, if we choose my suggestion, then not inherited parameters are not required to be refactored, but I'd encourage the effort for this as well :) Maybe it's even possible to mix both ([non-]inherited) params within one class...

shiramax:
thanks @ofedoren I will refactor it with the non-inherited params.
also, this PR can only be merged after theforeman/foreman#7434 will be merged.

PR #497, opened 2020-02-12
fixes #28973 - Add disable option for user

domitea:
Hammer part of https://github.com/theforeman/foreman/pull/7402

### DIFF
--- a/lib/hammer_cli_foreman/user.rb
+++ b/lib/hammer_cli_foreman/user.rb
@@ -12,6 +12,7 @@ module HammerCLIForeman
         field :full_name, _("Name")
         field :mail, _("Email")
         field :admin, _("Admin"), Fields::Boolean
+        field :disabled, _("Disabled"), Fields::Boolean
         field :last_login_on, _("Last login"), Fields::Date
         field nil, _("Authorized by"), Fields::SingleReference, :key => :auth_source
       end


lib/hammer_cli_foreman/param_filters.rb
```diff
@@ -25,6 +25,7 @@ def flatten_params(params)
       params.each do |p|
         result += flatten_params(p.params)
       end
+      puts result
```
shiramax: Is there a reason to print the result?

ofedoren: I think this is just debug leftover.

lib/hammer_cli_foreman/user.rb
```diff
@@ -57,6 +58,8 @@ def self.included(base)
         base.option '--default-location', 'DEFAULT_LOCATION_NAME', _("Default location name")
         base.option "--ask-password", "ASK_PW", " ",
                     :format => HammerCLI::Options::Normalizers::Bool.new
+        base.option '--disabled', 'USER_DISABLED', ' ',
+                    :format => HammerCLI::Options::Normalizers::Bool.new
```
ofedoren: Actually this is not needed unless there is no such param in API doc or you want to apply custom changes before sending the request.

The `--disabled` option will be created automatically after `hammer -r` if apidoc cache was updated.

lib/hammer_cli_foreman/user.rb
```diff
@@ -82,6 +85,12 @@ class UpdateCommand < HammerCLIForeman::UpdateCommand
       include CommonUpdateOptions
 
       build_options
+
+      def request_params
+        params = super
+        params['user']['disabled'] = option_disabled
+        params
+      end
```
ofedoren: This is not needed.

lib/hammer_cli_foreman/param_filters.rb
domitea: Yes, It was a leftover. Thanks!

domitea:
@ofedoren @shiramax Code improved, looking good for you?

ofedoren:
ok to test

shiramax:
no, thank you @domcleal  and @ofedoren, merging :)

PR #498, opened 2020-02-12
Fixes #22573 - Deprecate --root-pass in host group

ofedoren:


xprazak2:
The deprecation warning is printed for list command as well, is that expected?

deprecated-hammer

ofedoren:
@xprazak2, greate catch! I've updated PR so the deprecation warning is shown when using the option directly.

xprazak2:
Thanks!

PR #494, opened 2020-02-10
Fixes #28869 - change the description of the scoped loc and org

shiramax:

PR #503, opened 2020-02-20
Fixes #29096 - Remove duplicate API requests on AddAssociatedCommand

jturel:
#494 introduced a change which caused hammer-cli-katello tests to fail because of duplicate API requests. This PR cleans that up.

This should be cherry-picked onto the 2.0 branch as well

tail the app logs and watch the requests when running an 'add associated command' such as

```
hammer content-view add-version --id 1 --content-view-id 3 --content-view-version 2.1
```

the index method for the associated object is called twice

### DIFF
--- a/lib/hammer_cli_foreman/commands.rb
+++ b/lib/hammer_cli_foreman/commands.rb
@@ -671,8 +671,12 @@ module HammerCLIForeman
 
     def get_new_ids
       ids = get_current_ids.map(&:to_s)
-      required_ids = get_associated_identifiers.nil? ? [] : get_associated_identifiers.map(&:to_s)
-      required_ids << get_associated_identifier.to_s unless get_associated_identifier.nil?
+
+      associated_identifiers = get_associated_identifiers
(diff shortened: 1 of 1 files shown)


jturel:
@yifatmakias you may want to review this :)

ofedoren:
@jturel, thanks!

I wonder why I didn't notice multiple calls earlier... Our tests pass and I don't see any extra call.

Anyway, merging. The changes make sense, since there is no caching in `get_associated_identifier[s]`.
 
@yifatmakias, we probably need to apply those changes in https://github.com/theforeman/hammer-cli-foreman/pull/502 as well.

jturel:
Thanks @ofedoren !

I wrote the code this way since I wasn't sure if those get_* ,methods themselves should do the caching, or if it's expected that get_* should always issue an HTTP request.

jturel:
@ofedoren @shiramax how can we ensure this makes it into the 2.0-stable branch?

ofedoren:
@jturel feel free to open a CP PR if you want.

PR #508, opened 2020-03-09
Fixes #29274 - Use right param for os default template

ofedoren:
I guess this should be CPed into 2.0-stable.


tbrisker:
Please CP to 2.0-stable, reported by user at https://community.theforeman.org/t/hammer-os-set-default-template/17896/8

PR #509, opened 2020-03-12
Fixes #29326 - add mail notification command

yifatmakias:
Added mail-notification command to hammer.

### DIFF
--- a/test/functional/mail_notification_test.rb
+++ b/test/functional/mail_notification_test.rb
@@ -1,0 +1,57 @@
+require File.join(File.dirname(__FILE__), 'test_helper')
+
+describe 'mail_notification' do
+  describe 'list' do
+    before do
+      @cmd = %w[mail-notification list]
+      @mail_notifications = [{
+        id: 1,
+        name: 'config_summary',
+        description: 'A summary of eventful configuration management reports',
+        subscription_type: 'report'
+      }]
+    end
+
+    it 'should return a list of mail notifications' do
+      api_expects(:mail_notifications, :index, 'List mail notifications').returns(@mail_notifications)
+
+      output = IndexMatcher.new([
+                                  %w[ID NAME],
+                                  %w[1 config_summary]
+                                ])
+      expected_result = success_result(output)
+
+      result = run_cmd(@cmd)
+      assert_cmd(expected_result, result)
+    end
+  end
+
+  describe 'info' do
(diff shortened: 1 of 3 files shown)


test/functional/mail_notification_test.rb
```diff
@@ -0,0 +1,42 @@
+require File.join(File.dirname(__FILE__), 'test_helper')
+
```
ofedoren: :heart:

test/functional/mail_notification_test.rb
```diff
@@ -0,0 +1,42 @@
...
+        subscription_type: 'report'
+      }
+    end
+
+    it 'should return the info of a mail notification' do
+      params = ['--id=1']
+      api_expects(:mail_notifications, :show, 'Info mail notification').returns(@mail_notification)
+
+      result = run_cmd(@cmd + params)
+      result.exit_code.must_equal HammerCLI::EX_OK
```
ofedoren: Instead of checking for exit code, I'd rather check for the output as we do in: https://github.com/theforeman/hammer-cli-foreman/blob/master/test/functional/personal_access_token_test.rb#L91-L103

test/functional/mail_notification_test.rb
```diff
@@ -0,0 +1,42 @@
...
+        description: 'A summary of eventful configuration management reports',
+        subscription_type: 'report'
+      }]
+    end
+
+    it 'should return a list of mail notifications' do
+      api_expects(:mail_notifications, :index, 'List mail notifications').returns(@mail_notifications)
+
+      result = run_cmd(@cmd)
+      result.exit_code.must_equal HammerCLI::EX_OK
```
ofedoren: The same instead of checking for exit code, but for list commands: https://github.com/theforeman/hammer-cli-foreman/blob/master/test/functional/personal_access_token_test.rb#L43-L51

yifatmakias:
I fixed the tests, Thanks for the great review as always :)
regarding this comment you wrote:
It's a shame that we don't have API endpoints for actual manipulation (create/update/delete), but for viewing only :(

I am now working on adding to the API and to hammer the ability to add/remove mail notifications to/from a user. this is the create/update/delete that I think is missing. because it is not possible to change a mail notification or add one it is possible to assign it to a user.
It will be of course in a different PR.

PR #511, opened 2020-03-24
Add rake task with plugin template

ofedoren:
This adds a possibility to generate a simple draft for a new plugin via rake task `rake plugin:draft -- [options]`.

I'm not sure though whether this should be a rake task, hammer subcommand or hammer executable, or just directory with files...

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -4,6 +4,7 @@ require 'ci/reporter/rake/minitest'
 require 'hammer_cli/i18n/find_task'
 require_relative './lib/hammer_cli_foreman/version'
 require_relative './lib/hammer_cli_foreman/i18n'
+require_relative './lib/hammer_cli_foreman/task_helper'
 
 Rake::TestTask.new do |t|
   t.libs.push "lib"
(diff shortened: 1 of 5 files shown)


Rakefile
```diff
@@ -17,3 +17,160 @@ namespace :pkg do
...
+    require 'fileutils'
+
+    options = {
+      name: 'TODO', path: '../.', author: 'TODO', email: 'TODO'
+    }
+    OptionParser.new do |opts|
+      opts.banner = "Create a draft for a plugin.
+Example: rake plugin:draft -- --name=my_plugin --author='John Doe'\n
+Usage: rake plugin:draft -- [options]"
+      opts.on('-nNAME', '--name=NAME', "Plugin's name, e.g. my_plugin") { |name| options[:name] = name }
```
xprazak2, line 34: Having no separator between short flag and value seems very unusual to me. Could we have '=' to be consistent with `--name=NAME`?

xprazak2:
Very nice.

I am ok with the rake task, I think it is a good fit for this. Directory with files is not a bad idea, given that most of the content in the files does not need to be changed. At least that is what [foreman_plugin_template](https://github.com/theforeman/foreman_plugin_template) does, though I am not a fan of how they handle [renaming](https://github.com/theforeman/foreman_plugin_template/blob/master/rename.rb#L39). Maybe having placeholders in files that we know we need to replace? The rake task is a bit too long, might be worth splitting it up a bit.

Rakefile
```diff
@@ -17,3 +17,160 @@ namespace :pkg do
...
+    '#{plugin_namespace}::TodoCommand',
+    '#{plugin_name}/todo'
+  )
+end
+EOF
+      end
+      File.open("lib/#{plugin_name}/todo.rb", 'w') do |todo_command|
+        todo_command.write <<-EOF
+module #{plugin_namespace}
+  class TodoCommand < HammerCLIForeman::Command
```
xprazak2, line 127: I'd replace `Todo` with a more conventional noun to make it obvious that this is a resource/object/entity to be managed. And now for the hard part - to find a name for the example resource. Let the bikeshedding begin! :grin:

Rakefile
```diff
@@ -17,3 +17,160 @@ namespace :pkg do
...
 end
+
+namespace :plugin do
+  desc 'See rake plugin:draft -- --help for more information.'
+  task :draft do |_t, args|
+    require 'optparse'
+    require 'fileutils'
+
+    options = {
+      name: 'TODO', path: '../.', author: 'TODO', email: 'TODO'
```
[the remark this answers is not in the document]
ofedoren: Agree on that one, but as far as I know `OptionParser` doesn't have required options, so I've changed the logic. Now it's interactively prompted.

Rakefile
ofedoren: :) Well, I've decided to use general `Resource`, but I'm still open for suggestions :D

ofedoren:
@xprazak2, thanks for review! I've refactored code a bit so the task looks cleaner. Also changed the way of gathering the options.

lib/hammer_cli_foreman/task_helper.rb
```diff
@@ -0,0 +1,180 @@
...
+      def mk_gemspec
+        File.open("#{@name}.gemspec", 'w') do |gemspec|
+          gemspec.write <<-EOF
+lib = File.expand_path('../lib', __FILE__)
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+require '#{@name}/version'
+
+Gem::Specification.new do |spec|
+  spec.name          = '#{@name}'
+  spec.version       = #{@plugin_namespace}.version.dup
```
xprazak2, line 73: I think this should be:  `#{@plugin_namespace.version.dup}`

lib/hammer_cli_foreman/task_helper.rb
```diff
@@ -0,0 +1,180 @@
...
+
+  spec.platform      = Gem::Platform::RUBY
+  spec.summary       = 'Foreman #{@capitalized_bits.join(' ')} plugin for Hammer CLI'
+
+  # TODO: Don't forget to update required files accordingly!
+  spec.files         = Dir['{lib,config}/**/*', 'LICENSE', 'README*']
+  spec.require_paths = ['lib']
+  spec.test_files    = Dir['{test}/**/*']
+
+  spec.add_dependency 'hammer_cli_foreman', '>= 2.0.0'
```
xprazak2, line 87: Do we really want open-ended dependency?

xprazak2:
The interactive prompts are really good! The task is so much shorter and the new structure is so... structured.

lib/hammer_cli_foreman/task_helper.rb
ofedoren: Actually, if we do that, then `spec.version` will be hardcoded to the current one `0.0.1`.

But if we leave it in this way, it will be as we do in most plugins. So the new version is picked automatically when the gem is being built.

ofedoren:
@xprazak2, thanks for re-review :) I agree on open-ended dependency shouldn't be open, so I've changed that.

xprazak2:
Thanks!

PR #513, opened 2020-03-29
Fixes #7665 - added manage command to user mail notifications

yifatmakias:
This PR is blocked by another PR:
https://github.com/theforeman/foreman/pull/7549


yifatmakias:
@shiramax @ofedoren Now that the API changes are merged can you review this PR?

lib/hammer_cli_foreman/user_mail_notification.rb
```diff
@@ -0,0 +1,54 @@
+module HammerCLIForeman
+
+  class UserMailNotification < HammerCLIForeman::Command
+    resource :mail_notifications
+    command_name 'mail-notification'
+
+    class AddMailNotificationCommand < HammerCLIForeman::Command
```
ofedoren, line 7: This can be `HammerCLIForeman::CreateCommand`.

lib/hammer_cli_foreman/user_mail_notification.rb
```diff
@@ -0,0 +1,54 @@
...
+      action :create
+      command_name 'add'
+
+      success_message _('User mail notifications was added.')
+      failure_message _('Could not add user mail notification')
+
+      build_options
+    end
+
+    class UpdateMailNotificationCommand < HammerCLIForeman::Command
```
ofedoren, line 17: This can be `HammerCLIForeman::UpdateCommand`.

lib/hammer_cli_foreman/user_mail_notification.rb
```diff
@@ -0,0 +1,54 @@
...
+      action :update
+      command_name 'update'
+
+      success_message _('User mail notifications was updated.')
+      failure_message _('Could not update user mail notification')
+
+      build_options
+    end
+
+    class RemoveMailNotificationCommand < HammerCLIForeman::Command
```
ofedoren, line 27: This can be HammerCLIForeman::DeleteCommand.

lib/hammer_cli_foreman/user_mail_notification.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    class ListCommand < HammerCLIForeman::ListCommand
+      action :user_mail_notifications
+
+      output do
+        field :mail_notification_id, _('Id')
+        field :name, _('Name')
+        field :description, _('Description')
+        field :interval, _('Interval')
+        field :mail_query, _('Mail query') unless :mail_query.nil?
```
ofedoren, line 45: What exactly you want to achieve here?

`:mail_query.nil?` is always `false`.

test/functional/user_mail_notification_test.rb
```diff
@@ -0,0 +1,32 @@
+require File.join(File.dirname(__FILE__), 'test_helper')
+
```
ofedoren, line 2: :heart:

I'd also prefer to write just one test per PR, but if we've got updated test data and more than one command, then it will be nice to cover them as well.

lib/hammer_cli_foreman/user_mail_notification.rb
yifatmakias: Only with one type of mail notification, there is a mail query field so I thought this way it will show the field only when it is relevant. In a second check, the column will appear all the time, if there is a mail query it will appear in the column, if not it will be empty.
long story short - I removed the unless part it was redundant.

test/functional/user_mail_notification_test.rb
yifatmakias: I added more tests to all of the commands. You are definitely right there should be at least one test to each command.

lib/hammer_cli_foreman/user_mail_notification.rb
```diff
@@ -0,0 +1,54 @@
+module HammerCLIForeman
+
+  class UserMailNotification < HammerCLIForeman::Command
+    resource :mail_notifications
+    command_name 'mail-notification'
+
+    class AddMailNotificationCommand < HammerCLIForeman::CreateCommand
+      action :create
```
ofedoren, line 8: Sorry, forgot to mention: if you inherit from `[Create|Update|Delete|Info|List]Command` you don't need to specify the `action` _unless_ it's different from default.

lib/hammer_cli_foreman/user_mail_notification.rb
```diff
@@ -0,0 +1,54 @@
...
+      command_name 'add'
+
+      success_message _('User mail notifications was added.')
+      failure_message _('Could not add user mail notification')
+
+      build_options
+    end
+
+    class UpdateMailNotificationCommand < HammerCLIForeman::UpdateCommand
+      action :update
```
ofedoren, line 18: same here

lib/hammer_cli_foreman/user_mail_notification.rb
```diff
@@ -0,0 +1,54 @@
...
+      command_name 'update'
+
+      success_message _('User mail notifications was updated.')
+      failure_message _('Could not update user mail notification')
+
+      build_options
+    end
+
+    class RemoveMailNotificationCommand < HammerCLIForeman::DeleteCommand
+      action :destroy
```
ofedoren, line 28: same here

lib/hammer_cli_foreman/user_mail_notification.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    class ListCommand < HammerCLIForeman::ListCommand
+      action :user_mail_notifications
+
+      output do
+        field :mail_notification_id, _('Id')
+        field :name, _('Name')
+        field :description, _('Description')
+        field :interval, _('Interval')
+        field :mail_query, _('Mail query')
```
ofedoren, line 45: Actually, if you want to hide `mail_query` field you can specify `hide_blank` option:
`field :mail_query, _('Mail query'), Fields::Field, hide_blank: true`. This will hide the field if value for it is `nil`. Although it doesn't work for table output. So, it's up to you whether have it or not.

yifatmakias: I don't see a reason to add this if it does not work for table output. Do you?

ofedoren: Well, it works for base, csv and other outputs, which are used more for automation purposes. I don't see much usage of this option in the project anyway, so we can leave it as it is now.

yifatmakias: Sounds good to me :)

ofedoren:
Thanks, @yifatmakias! All seems to be ready for merge, so merging this one.

PR #515, opened 2020-04-20
[WIP] Fixes #29575 - Session after expiration considers the present auth type

rahulbajaj0509:
With the changes merged related to SSO, the session is created properly, contains the auth type but when I run some hammer command it tries to login with basic auth and asks for username as it was ignoring the session at all.

With the patch, there is still a issue. For the first time when user session expires, hammer asks for username and password. While from the second time it checks for the Auth type in the session and gives correct output.


rahulbajaj0509:
This was realized by @mbacovsky back when we were working the initial SSO pull request. Look here: https://github.com/theforeman/hammer-cli-foreman/pull/405#pullrequestreview-306894840

PR #516, opened 2020-04-21
[WIP] Fixes #29575 - fix session expiry of oauth users

rahulbajaj0509:
This PR is not at all review ready, I just want to make sure I am on the right path and can make changes to the `foreman_api_connection` method. This PR is related to https://github.com/theforeman/hammer-cli-foreman/pull/515/files. 

What do you think about this @ofedoren @mbacovsky

### DIFF
--- a/lib/hammer_cli_foreman/api/connection.rb
+++ b/lib/hammer_cli_foreman/api/connection.rb
@@ -52,9 +52,8 @@ module HammerCLIForeman
         return AUTH_TYPES[:basic_auth] unless HammerCLIForeman::Sessions.enabled?
 
         url = settings.get(:_params, :host) || settings.get(:foreman, :host)
-        username = settings.get(:_params, :username) || settings.get(:foreman, :username)
         session = HammerCLIForeman::Sessions.get(url)
-        if !session.valid? && session.user_name == username && !session.auth_type.nil?
+        if !session.valid? && !session.auth_type.nil?
           session.auth_type
         else
           # If the caller has not sepcified an 'auth_type'
@@ -92,7 +91,21 @@ module HammerCLIForeman
 
   def self.foreman_api_connection
     HammerCLI.context[:api_connection].create(CONNECTION_NAME) do
-      HammerCLIForeman::Api::Connection.new(HammerCLI::Settings, Logging.logger['API'], HammerCLI::I18n.locale)
+      settings = HammerCLI::Settings
+      if settings.get(:foreman, :use_sessions) == true
+        url = settings.get(:_params, :host) || settings.get(:foreman, :host)
+        session = HammerCLIForeman::Sessions.get(url)
+        if session.valid? && session.expired?
+          HammerCLIForeman::Api::Connection.new(settings, Logging.logger['API'],
+            HammerCLI::I18n.locale, session.auth_type)
+        else
+          HammerCLIForeman::Api::Connection.new(settings, Logging.logger['API'],
+            HammerCLI::I18n.locale)
+        end
+      else
+        HammerCLIForeman::Api::Connection.new(settings, Logging.logger['API'],
+        HammerCLI::I18n.locale)
+      end
     end
   end
 
--- a/lib/hammer_cli_foreman/api/connection.rb
+++ b/lib/hammer_cli_foreman/api/connection.rb
@@ -91,7 +91,7 @@ module HammerCLIForeman
 
   def self.foreman_api_connection
     HammerCLI.context[:api_connection].create(CONNECTION_NAME) do
-      if session_auth = previous_session_auth(HammerCLI::Settings)
+      if session_auth = previous_session_auth
         HammerCLIForeman::Api::Connection.new(HammerCLI::Settings, Logging.logger['API'],
           HammerCLI::I18n.locale, session_auth)
       else
@@ -112,7 +112,8 @@ module HammerCLIForeman
     end
   end
 
-  def self.previous_session_auth(settings)
+  def self.previous_session_auth
+    settings = HammerCLI::Settings
     return nil unless settings.get(:foreman, :use_sessions)
 
     url = settings.get(:_params, :host) || settings.get(:foreman, :host)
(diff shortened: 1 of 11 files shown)


rahulbajaj0509:
@ofedoren @mbacovsky this PR is ready for testing. I am actively fixing test cases and adding new test cases to the PR. Let me know how this PR looks code wise.

lib/hammer_cli_foreman/api/connection.rb
```diff
@@ -52,9 +52,8 @@ def default_auth_type(settings)
         return AUTH_TYPES[:basic_auth] unless HammerCLIForeman::Sessions.enabled?
 
         url = settings.get(:_params, :host) || settings.get(:foreman, :host)
-        username = settings.get(:_params, :username) || settings.get(:foreman, :username)
         session = HammerCLIForeman::Sessions.get(url)
-        if !session.valid? && session.user_name == username && !session.auth_type.nil?
```
ofedoren, line 57: I think we had that for this case:
 - I'm user `A`
 - I was logged in with `code grant auth`
 - I logged out
 - I want to log in again
 - I assume `code grant auth` again.

If we change that, then user `B` can be prompted for `code` even though he wanted `password grant auth`.

What do you think?

lib/hammer_cli_foreman/api/connection.rb
```diff
@@ -107,6 +112,14 @@ def self.foreman_api_reconnect(auth_type)
     end
   end
 
+  def self.previous_session_auth(settings)
+    return nil unless settings.get(:foreman, :use_sessions)
+
+    url = settings.get(:_params, :host) || settings.get(:foreman, :host)
+    session = HammerCLIForeman::Sessions.get(url)
+    session.auth_type if session.valid? && session.expired?
+  end
```
ofedoren, line 121: This method is used only here, so we can change it to
```ruby
def self.previous_session_auth
  settings = HammerCLI::Settings
  #... same code
end
```

If you agree, don't forget to adjust line 94.

lib/hammer_cli_foreman/sessions.rb
```diff
@@ -88,5 +91,17 @@ def destroy
...
+      @auth_type == AUTH_TYPES[:oauth_authentication_code_grant] ||
+        @auth_type == AUTH_TYPES[:oauth_password_grant]
+    end
+
+    def expired?
+      return false unless auth_type_oauth?
+      return false unless (Time.parse(token_expires_at) - Time.now.utc).to_i.negative?
+
+      true
+    end
```
ofedoren, line 105: How about:
```ruby
def expired?
  return false unless auth_type_oauth?
  return true if token_expires_at.to_s.empty? # we don't support empty expiration dates or assume the expiration date is  now
  return false unless (Time.parse(token_expires_at.to_s).utc - Time.now.utc).to_i.negative? # convert all time to UTC and then compare

  true
end
```

lib/hammer_cli_foreman/api/connection.rb
rahulbajaj0509: When made this change I thought of this use case:

* I'm a user `A`
* In my config file, by default i have the setting - `username` set to `admin`.
* I login with user `A`, using the `code grant auth`
* Now, my session expires, and it checks `session.user_name == username`, where my `session.username` is `A` but my `username` is `admin`
* So it goes in the else part and takes `basic auth` 

Let me know your thoughts.

rahulbajaj0509: Giving it a second thought, if in the case described, the user wants to relogin with a different auth source. If they pass in the command from hammer shell, then they will be redirected to the `foreman_api_reconnect` and correct auth type would be set. 

Let me know your thoughts. I think this needs a little brainstorming though :+1:

test/unit/connection_test.rb
```diff
@@ -0,0 +1,96 @@
...
+      HammerCLI::Connection.any_instance.expects(:drop).with(HammerCLIForeman::CONNECTION_NAME)
+      subject
+    end
+
+    it 'creates a new connection to foreman' do
+      HammerCLI::Connection.any_instance.expects(:create).with(HammerCLIForeman::CONNECTION_NAME)
+      subject
+    end
+  end
+end
```
rahulbajaj0509, line 24: @ofedoren can you please take a look at these test cases? They seem to pass in the file but all the other files test cases are failing due to this change. I think it is mostly because of the the `:drop` method. Once dropped it is unable to create a connection again.

rahulbajaj0509: Like `before` block I tried to use the `after` block too to create a connection again but it does not seem to be working that way :(

lib/hammer_cli_foreman/sessions.rb
rahulbajaj0509: I think the second condition here, `we don't support empty expiration dates or assume the expiration date is  now` gets handled by the third condition. I mean if the token_expires_at is not present then in the third condition the `(Time.parse(token_expires_at.to_s).utc - Time.now.utc).to_i.negative?` will be true. Finally, the function will return true. 

Let me know if still think the condition must be added?

test/unit/connection_test.rb
ofedoren: Well.. I think i found why is that:

`HammerCLIForeman.foreman_api_reconnect` creates an actual connection while in tests we create a fake connection:

https://github.com/theforeman/hammer-cli-foreman/blob/master/test/test_helper.rb#L23-L25
and
https://github.com/theforeman/hammer-cli-foreman/blob/637846b3d856b9683657b780c945ae678085b027/lib/hammer_cli_foreman/testing/api_expectations.rb#L153-L161

rahulbajaj0509:
We had a call and the summary is:
* We have two sources to determine the auth_source: 
  * Hammer configuration file 
  * Session file
* If we consider referring to the `session file` then we need to remove the condition `session.user_name == username ` as explained here: https://github.com/theforeman/hammer-cli-foreman/pull/516#discussion_r420170358. But with a change in the `Hammer configuration file` causes error as explained here: https://github.com/theforeman/hammer-cli-foreman/pull/516#discussion_r420170358.
* Figured out we can not have two sources of gathering the auth source information. We need only one source for determining the value of auth source. The idea is that `Hmmer configuration file` should be the source for the correct auth source.

To summerize the user-story:
*  User can login in hammer using the `hammer auth login oauth` or `hammer auth login basic`. Once user session expires, hammer will try to login with the `default_auth_type` configured in the `Hammer configuration file`.
* If no `default_auth_type` is present, then hammer will try to login with the`basic auth`

PR #517, opened 2020-05-01
Fixes #29635 - Ask for oauth code only when needed

ofedoren:
~~Requires https://github.com/theforeman/hammer-cli-foreman/pull/516. Please don't merge it right away  :)~~

~~Includes commits from the mentioned PR. The fix is in the last commit.~~

@rahulbajaj0509, does it fix the issue? ~~If so, we could include this in your PR and resolve two issues by one PR since my solution is based on your changes and I don't see the reason to cut something from yours and add to mine :)~~

### DIFF
--- a/lib/hammer_cli_foreman/auth.rb
+++ b/lib/hammer_cli_foreman/auth.rb
@@ -46,16 +46,16 @@ module HammerCLIForeman
           if option_two_factor?
             Oauth.execute_with_params(
               AUTH_TYPES[:oauth_authentication_code_grant],
-              option_oidc_token_endpoint,
-              option_oidc_authorization_endpoint,
-              option_oidc_client_id,
-              option_oidc_redirect_uri
+              option_oidc_token_endpoint || HammerCLI::Settings.get(:foreman, :oidc_token_endpoint),
+              option_oidc_authorization_endpoint || HammerCLI::Settings.get(:foreman, :oidc_authorization_endpoint),
+              option_oidc_client_id || HammerCLI::Settings.get(:foreman, :oidc_client_id),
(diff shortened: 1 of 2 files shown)


ofedoren:
@rahulbajaj0509 updated. Now it doesn't require your changes and can be merged alongside.

lib/hammer_cli_foreman/auth.rb
```diff
@@ -46,16 +46,16 @@ def execute
...
             Oauth.execute_with_params(
               AUTH_TYPES[:oauth_authentication_code_grant],
-              option_oidc_token_endpoint,
-              option_oidc_authorization_endpoint,
-              option_oidc_client_id,
-              option_oidc_redirect_uri
+              option_oidc_token_endpoint || HammerCLI::Settings.get(:foreman, :oidc_token_endpoint),
+              option_oidc_authorization_endpoint || HammerCLI::Settings.get(:foreman, :oidc_authorization_endpoint),
+              option_oidc_client_id || HammerCLI::Settings.get(:foreman, :oidc_client_id),
+              option_oidc_redirect_uri || HammerCLI::Settings.get(:foreman, :oidc_redirect_uri)
```
rahulbajaj0509, line 52: I am not sure if this should be passed here. I want to make sure this change does not have any side effects. Will further test and let you know.

rahulbajaj0509: BTW, a lot of the testing also depends upon this conversation: https://github.com/theforeman/hammer-cli-foreman/pull/516/files#r420009404 lets resolve this conversation and make a decision.

shiramax:
thanks @ofedoren and thank you @rahulbajaj0509 for the review!

PR #528, opened 2020-06-16
Fixes #30130 - display request UUID for audit

ares:


lib/hammer_cli_foreman/audit.rb
```diff
@@ -61,6 +61,7 @@ def extend_data(audit)
           field :old, _("Old"), Fields::LongText, :hide_blank => true
           field :new, _("New"), Fields::LongText, :hide_blank => true
         end
+        field :request_uuid, _("Request UUID")
```
ofedoren, line 64: I think this field can be moved in the `ListCommand` output definition, but with field sets specified like so:
```ruby
field :request_uuid, _("Request UUID"), Fields::Field, sets: ['ALL']
```
to be shown only by explicit request with `--fields 'ALL'` option.

lib/hammer_cli_foreman/audit.rb
```diff
@@ -61,6 +61,7 @@ def extend_data(audit)
...
       end
```
ofedoren, line 66: Right away after the changes in previous comment, we can use
```ruby
extend_output_definition do |definition|
  definition.insert(:replace, :request_uuid) do
    field :request_uuid, _("Request UUID"), Fields::Field, sets: ['ALL', 'DEFAULT']
  end
end
```
here, so the `Request UUID` field is shown by default.

ares:
Thanks @ofedoren, updated

PR #534, opened 2020-08-19
Fixes #30663 - add ptables and realms to location and organization info

yifatmakias:


ofedoren (approved):
Thanks, @yifatmakias! Works as expected.

PR #536, opened 2020-08-27
Fixes #30720 - use underscores in `login oauth` option values

mirekdlugosz:
Not tested, but that's the only place where `OPENIDC-TOKEN-ENDPOINT` appears, so should to be very safe change.

If I wanted this change to be included in Satellite 6.8, I would need to cherry-pick to which branch?
And would I need to do something beside cherry-picking?


ofedoren (approved):
Thanks, @mirekdlugosz! The changes are safe, so merging :)

If you want to do a cherry-pick, please open a PR against `2.1-stable`. After the PR is merged, we will do a new release and packaging.

mirekdlugosz:
@ofedoren I created #538 with cherry-pick. Not super-important, but would be nice to have that in Sat 6.8 to avoid inconsistency in hammer :) .

Thanks for guidance and merge!

PR #541, opened 2020-09-13
Fixes #30805 - added missing tests to installation medium test

yifatmakias:


test/functional/media_test.rb
```diff
@@ -0,0 +1,130 @@
...
+        path: 'http://mirror.centos.org/centos/$major/os/$arch',
+        os_family: 'Redhat',
+        operating_systems: [],
+        locations: [],
+        organizations: []
+      }]
+    end
+
+    it 'should return a list of mediums' do
+      api_expects(:media, :index, 'List mediums').returns(@mediums)
```
xprazak2, line 19: nit: do you think it would be reasonable to use `media` as a plural so that we are consistent? 

Also one line above in test name.

xprazak2:
Looks good, I have just one nitpick.

xprazak2:
Thanks!

PR #543, opened 2020-09-21
Fixes #30886 - added missing tests to setting

yifatmakias:
The setting command in hammer was missing the show and update tests.


xprazak2:
Thanks!

PR #545, opened 2020-09-29
Fixes #30496 - Return non-zero exit code if services failed

ofedoren:


xprazak2:
Works as expected, will there be a follow-up for Katello as well?

ofedoren:
> Works as expected, will there be a follow-up for Katello as well?

Thanks :) I guess there is no need to do something with Katello plugin. This fix just makes behaviour of the new default ping command the same as the old one (if something failed, return 1).

ofedoren:
@jturel, since you've reported the issue, could you please take a look and confirm if this fixes the issue? :)

shiramax (approved):
thanks @ofedoren code wise it looks good!

jturel (approved):
ACK!

I got return code 1 after I stopped tomcat (candlepin) in my katello env. Just what I would expect. Thank you @ofedoren !

xprazak2:
Thanks everyone!

ekohl:
On a nightly box I get:
```
[root@pipeline-katello-server-nightly-centos7 bats_results]# hammer ping
database:         
    Status:          ok
    Server Response: Duration: 0ms
candlepin:        
    Status:          ok
    Server Response: Duration: 22ms
candlepin_events: 
    Status:          ok
    message:         11 Processed, 0 Failed
    Server Response: Duration: 0ms
candlepin_auth:   
    Status:          ok
    Server Response: Duration: 41ms
katello_events:   
    Status:          ok
    message:         4 Processed, 0 Failed
    Server Response: Duration: 0ms
pulp:             
    Status:          ok
    Server Response: Duration: 64ms
pulp_auth:        
    Status:          ok
    Server Response: Duration: 26ms
foreman_tasks:    
    Status:          ok
    Server Response: Duration: 2ms

[root@pipeline-katello-server-nightly-centos7 bats_results]# echo $?
1
```

I think it's because Pulp 3 failed on my setup but this isn't clear to me as a user.

ofedoren:
> On a nightly box I get:
> 
> ```
> [root@pipeline-katello-server-nightly-centos7 bats_results]# hammer ping
> database:         
>     Status:          ok
>     Server Response: Duration: 0ms
> candlepin:        
>     Status:          ok
>     Server Response: Duration: 22ms
> candlepin_events: 
>     Status:          ok
>     message:         11 Processed, 0 Failed
>     Server Response: Duration: 0ms
> candlepin_auth:   
>     Status:          ok
>     Server Response: Duration: 41ms
> katello_events:   
>     Status:          ok
>     message:         4 Processed, 0 Failed
>     Server Response: Duration: 0ms
> pulp:             
>     Status:          ok
>     Server Response: Duration: 64ms
> pulp_auth:        
>     Status:          ok
>     Server Response: Duration: 26ms
> foreman_tasks:    
>     Status:          ok
>     Server Response: Duration: 2ms
> 
> [root@pipeline-katello-server-nightly-centos7 bats_results]# echo $?
> 1
> ```
> 
> I think it's because Pulp 3 failed on my setup but this isn't clear to me as a user.

I'd think that the server response contains more services then hammer shows (probably due to newer API and outdated one of hammer plugins, e.g. h-c-katello).

ekohl:
I can indeed confirm that after I fixed my Pulpcore setup, it passed again. To me as a user it was confusing to see everything Status ok but still fail.

As a user I would think it either needs to pass if all the visible ones are ok or it should print a warning that there is another unlisted service that failed. Should I open a new issue or will you pick it up as part of 30496?

As for listing Pulpcore, I'll reach out to the Katello devs.

ofedoren:
> I can indeed confirm that after I fixed my Pulpcore setup, it passed again. To me as a user it was confusing to see everything Status ok but still fail.
> 
> As a user I would think it either needs to pass if all the visible ones are ok or it should print a warning that there is another unlisted service that failed. Should I open a new issue or will you pick it up as part of 30496?
> 
> As for listing Pulpcore, I'll reach out to the Katello devs.

Thanks for being concerned! :)

Well, I see two options:
 1. Be simple: hammer commands are just UI for API calls with extra steps, so if the server responds with something that hammer doesn't show, then behave like the additional stuff doesn't exist.
 2. Be consistent and simple: Consider this as a bug and poke devs of API part to poke hammer devs to add support for new services which aren't covered in hammer.

I'd go with 2. Because if we go with 1, user doen't get anything to be aware if something went wrong even tough in debug mode hammer shows the whole server response, where a user could find something that might help.

The approach you've suggested is like 1.5, which I'd like to avoid if you don't insist :(

P.S. I've just checked the latests and I see that there's missing `pulp3` service in the output...

ekohl:
It can also happen if the user has katello and only hammer-cli-foreman (not h-c-katello) installed. Granted, perhaps a bit of an edge case but a generalized version of that is that if you see any service that's not recognized, a warning is emitted to stderr (besides poking the Katello devs to support Pulp 3).

PR #547, opened 2020-09-30
Adding download policy to info fields

rafaelguerra01:
hammer proxy info --id <id> now shows the download policy for the given smart proxy.


yifatmakias (commented):
Thanks for your contribution :)
One change is needed - the attribute download_policy is coming from katello therefor this change should be added to hammer-cli-katello and not to hammer-cli-foreman

yifatmakias:
@rafaelguerra01 Please close this PR and open another one in hammer-cli-foreman with this change as I explained above.
Thanks! :)

PR #548, opened 2020-10-13
Fixes #32265 - Support unrecognized services in ping

ofedoren:
@ekohl, probably you want to review/test this as well.

This should add a warning message in case if a service failed and is not listed in the output.

ScreenShot-1602617557330

### DIFF
--- a/lib/hammer_cli_foreman/command_extensions/ping.rb
+++ b/lib/hammer_cli_foreman/command_extensions/ping.rb
@@ -10,9 +10,28 @@ module HammerCLIForeman
         end
       end
 
+      def self.check_for_unrecognized(plugins, output_definition)
+        failed = plugins.select { |_, data| data['services'] }
+                        .each_with_object([]) { |(_, d), s| s << d['services'] }
+                        .reduce({}, :merge)
+                        .select do |name, data|
+          begin
+            output_definition.find_field(name)
+            false
+          rescue ArgumentError
+            data['status'] == _('FAIL')
+          end
+        end
+        return if failed.empty?
+
+        warn [_('%{count} more service(s) failed, but not shown:') % { count: failed.size },
+              failed.keys.join(', '),
+              ''].join("\n")
+      end
+
       def self.failed?(services)
-        services.each_value.any? { |s| s['status'] == _('FAIL') } ||
-          services['foreman']['database']['active'] == 'FAIL'
+        services['foreman']['database']['active'] == 'FAIL' ||
+          services.each_value.any? { |s| s['status'] == _('FAIL') }
       end
 
       request_options do |options|
--- a/lib/hammer_cli_foreman/ping.rb
+++ b/lib/hammer_cli_foreman/ping.rb
@@ -19,7 +19,12 @@ module HammerCLIForeman
         response = send_request
         print_data(response)
 
-        return 1 if HammerCLIForeman::CommandExtensions::Ping.failed?(response)
+        if HammerCLIForeman::CommandExtensions::Ping.failed?(response)
+          HammerCLIForeman::CommandExtensions::Ping.check_for_unrecognized(
+            response, output_definition
+          )
+          return 1
+        end
 
         HammerCLI::EX_OK
       end
--- a/test/functional/ping_test.rb
+++ b/test/functional/ping_test.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require File.join(File.dirname(__FILE__), 'test_helper')
 
 describe 'ping' do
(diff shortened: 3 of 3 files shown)


lib/hammer_cli_foreman/command_extensions/ping.rb
```diff
@@ -10,9 +10,28 @@ class Ping < HammerCLI::CommandExtensions
...
+        warn [_('%{count} more service(s) failed, but not shown:') % { count: failed.size },
+              failed.keys.join(', '),
+              ''].join("\n")
+      end
+
       def self.failed?(services)
-        services.each_value.any? { |s| s['status'] == _('FAIL') } ||
-          services['foreman']['database']['active'] == 'FAIL'
+        services['foreman']['database']['active'] == 'FAIL' ||
+          services.each_value.any? { |s| s['status'] == _('FAIL') }
```
ekohl, line 34: Is there anything changing here? It looks like you're just swapping the order.

lib/hammer_cli_foreman/command_extensions/ping.rb
```diff
@@ -10,9 +10,28 @@ class Ping < HammerCLI::CommandExtensions
...
+                        .each_with_object([]) { |(_, d), s| s << d['services'] }
+                        .reduce({}, :merge)
+                        .select do |name, data|
+          begin
+            output_definition.find_field(name)
+            false
+          rescue ArgumentError
+            data['status'] == _('FAIL')
+          end
+        end
```
ekohl, line 24: This logic is quite hard to follow. Why does it modify data in place? Shouldn't it be a more pure function that returns a list of unrecognized services?

lib/hammer_cli_foreman/ping.rb
```diff
@@ -19,9 +19,12 @@ def execute
         response = send_request
         print_data(response)
 
-        return 1 if HammerCLIForeman::CommandExtensions::Ping.failed?(response)
+        return HammerCLI::EX_OK unless HammerCLIForeman::CommandExtensions::Ping.failed?(response)
 
-        HammerCLI::EX_OK
```
ekohl, line 24: I like the exit if it's abnormal and closing with `EX_OK`. That implies `EX_OK` is the expected execution. Your version implies exiting with 1 is normal.

lib/hammer_cli_foreman/command_extensions/ping.rb
ofedoren: yeap, exactly. That is just to "short-circuit" checks.

lib/hammer_cli_foreman/command_extensions/ping.rb
ofedoren: This method doesn't modify anything. If there is a failed service that has no field (won't be shown) it simply puts a warning. Otherwise does nothing additional.

lib/hammer_cli_foreman/ping.rb
ofedoren: Hmm... I thought that exit with `EX_OK` `unless` something failed should be done right away and don't check for which services will be shown and which won't.

I can refactor that, but I'm not sure if it will be more readable though..

ekohl: When I read a function (and this was also what previously happened) is that I'd expect all steps in between would error out early.

So roughly:
```ruby
return 1 if failure_condition_1
return 2 if failure_condition_2

# Everything went OK

0
```

This makes it easier to extend with more failure conditions. In this case, I'd suggest:
```ruby
if HammerCLIForeman::CommandExtensions::Ping.failed?(response)
  HammerCLIForeman::CommandExtensions::Ping.check_for_unrecognized(
     response, output_definition
  )
  return 1
end

HammerCLI::EX_OK
```

IMHO the flow is easier to follow: we see a failure, do some additional processing and return early. Does that make sense or am I explaining it poorly?

ofedoren: @ekohl, I just was going to update the PR :D

Yes, that makes sense.

ofedoren:
@ekohl, updated with your suggestion.

ekohl (approved):
Thanks!

ekohl:
@ofedoren I have merge permissions because I'm an admin, not because I'm a regular maintainer. Do you expect another hammer maintainer to review or shall I hit the big green button?

ofedoren:
@ekohl, well let's try to ask @yifatmakias :) Yifat, if you are interested or have time, could you please take a look?

yifatmakias:
Looks good to me. Thanks @ofedoren :)

PR #549, opened 2020-10-15
Fixes #31074 - added missing tests to filter

yifatmakias:
The filter command in hammer was missing the index, create, and destroy tests

### DIFF
--- a/test/functional/filter_test.rb
+++ b/test/functional/filter_test.rb
@@ -1,18 +1,21 @@
+# frozen_string_literal: true
+
 require File.join(File.dirname(__FILE__), 'test_helper')
 
 describe 'filter' do
   def api_expects_filter_info(options)
     override = !!options[:override]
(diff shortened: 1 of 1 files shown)


test/functional/filter_test.rb
```diff
@@ -21,7 +24,7 @@ def assert_update_success(result)
 
   describe 'create' do
```
ofedoren, line 25: Thanks, @yifatmakias! +1 for using default rubocop :rose:.

Could you please also add a test for `create` action which actually tries to create a filter? The one we have there uses `api_expects_no_call` and misses a pair.

yifatmakias: Yes, I will add another test for create :)

ofedoren (approved):
Thanks, @yifatmakias, LGTM :)

Merging this one, although I've noticed that we use pretty outdated `test/data`...

PR #550, opened 2020-10-23
Fixes #31140 - hammer ping skip authentication

ranjan:


ofedoren (commented):
Thanks, @ranjan!

Could you please move this code here: https://github.com/theforeman/hammer-cli-foreman/blob/master/lib/hammer_cli_foreman/command_extensions/ping.rb?

It should look like this:
```ruby
      request_options do |options|
        options[:with_authentication] = false
      end
```

ranjan:
> Thanks, @ranjan!
> 
> Could you please move this code here: https://github.com/theforeman/hammer-cli-foreman/blob/master/lib/hammer_cli_foreman/command_extensions/ping.rb?
> 
> It should look like this:
> 
> ```ruby
>       request_options do |options|
>         options[:with_authentication] = false
>       end
> ```

Thanks, @ofedoren PR updated.

ofedoren (approved):
Thanks, @ranjan! Works as expected.

PR #551, opened 2020-12-07
Fixes #31384 - fix hammer list failure when defaults are set

yifatmakias:
This fix is for a few commands that were failing whenever running list
if defaults were set before.

### DIFF
--- a/lib/hammer_cli_foreman/architecture.rb
+++ b/lib/hammer_cli_foreman/architecture.rb
@@ -11,7 +11,7 @@ module HammerCLIForeman
         field :name, _("Name")
       end
 
-      build_options
+      build_options expand: { except: %i[organizations locations] }, without: %i[organization_id location_id]
     end
 
 
@@ -23,7 +23,7 @@ module HammerCLIForeman
         HammerCLIForeman::References.timestamps(self)
       end
 
-      build_options
+      build_options expand: { except: %i[organizations locations] }, without: %i[organization_id location_id]
     end
 
 
@@ -31,7 +31,7 @@ module HammerCLIForeman
       success_message _("Architecture created.")
       failure_message _("Could not create the architecture")
 
-      build_options
+      build_options expand: { except: %i[organizations locations] }, without: %i[organization_id location_id]
     end
 
 
@@ -39,7 +39,7 @@ module HammerCLIForeman
       success_message _("Architecture deleted.")
       failure_message _("Could not delete the architecture")
 
-      build_options
+      build_options expand: { except: %i[organizations locations] }, without: %i[organization_id location_id]
     end
 
 
@@ -47,7 +47,7 @@ module HammerCLIForeman
       success_message _("Architecture updated.")
       failure_message _("Could not update the architecture")
 
-      build_options
+      build_options expand: { except: %i[organizations locations] }, without: %i[organization_id location_id]
     end
 
     HammerCLIForeman::AssociatingCommands::OperatingSystem.extend_command(self)
--- a/lib/hammer_cli_foreman/bookmark.rb
+++ b/lib/hammer_cli_foreman/bookmark.rb
@@ -15,34 +15,34 @@ module HammerCLIForeman
         field :owner_type, _('Owner Type')
       end
 
-      build_options
+      build_options expand: { except: %i[organizations locations] }, without: %i[organization_id location_id]
     end
 
     class InfoCommand < HammerCLIForeman::InfoCommand
       output ListCommand.output_definition
-      
-      build_options
+
+      build_options expand: { except: %i[organizations locations] }, without: %i[organization_id location_id]
     end
 
     class CreateCommand < HammerCLIForeman::CreateCommand
       success_message _('Bookmark %<name>s created.')
       failure_message _('Failed to create %<name>s bookmark')
 
-      build_options
+      build_options expand: { except: %i[organizations locations] }, without: %i[organization_id location_id]
     end
 
     class UpdateCommand < HammerCLIForeman::UpdateCommand
       success_message _('Bookmark %<name>s updated successfully.')
       failure_message _('Failed to update %<name>s bookmark')
 
-      build_options
+      build_options expand: { except: %i[organizations locations] }, without: %i[organization_id location_id]
     end
 
     class DeleteCommand < HammerCLIForeman::DeleteCommand
       success_message _('Bookmark deleted successfully.')
       failure_message _('Failed to delete bookmark')
 
-      build_options
+      build_options expand: { except: %i[organizations locations] }, without: %i[organization_id location_id]
     end
 
     autoload_subcommands
--- a/lib/hammer_cli_foreman/operating_system.rb
+++ b/lib/hammer_cli_foreman/operating_system.rb
@@ -13,7 +13,7 @@ module HammerCLIForeman
         field :family, _("Family")
       end
 
-      build_options
+      build_options expand: { except: %i[organizations locations] }, without: %i[organization_id location_id]
     end
 
 
@@ -36,7 +36,7 @@ module HammerCLIForeman
         HammerCLIForeman::References.parameters(self)
       end
 
-      build_options
+      build_options expand: { except: %i[organizations locations] }, without: %i[organization_id location_id]
     end
 
 
@@ -44,7 +44,7 @@ module HammerCLIForeman
       success_message _("Operating system created.")
       failure_message _("Could not create the operating system")
 
-      build_options
+      build_options expand: { except: %i[organizations locations] }, without: %i[organization_id location_id]
     end
 
 
@@ -52,7 +52,7 @@ module HammerCLIForeman
       success_message _("Operating system updated.")
       failure_message _("Could not update the operating system")
 
-      build_options
+      build_options expand: { except: %i[organizations locations] }, without: %i[organization_id location_id]
     end
 
 
@@ -60,7 +60,7 @@ module HammerCLIForeman
       success_message _("Operating system deleted.")
       failure_message _("Could not delete the operating system")
 
-      build_options
+      build_options expand: { except: %i[organizations locations] }, without: %i[organization_id location_id]
     end
 
 
@@ -76,7 +76,7 @@ module HammerCLIForeman
         validator.any(:option_operatingsystem_id, :option_operatingsystem_title).required
       end
 
-      build_options
+      build_options expand: { except: %i[organizations locations] }, without: %i[organization_id location_id]
     end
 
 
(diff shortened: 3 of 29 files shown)


lib/hammer_cli_foreman/architecture.rb
```diff
@@ -11,6 +11,12 @@ class ListCommand < HammerCLIForeman::ListCommand
         field :name, _("Name")
       end
 
+      def option_sources
+        sources = super
+        context[:defaults]&.defaults_settings&.clear
```
ofedoren, line 16: This can have unpleasant consequences:
 - What if default settings contain not only taxonomy related options? `.clear` will wipe all of it.
 - Try to set defaults then run `hammer shell` > `defaults list` > `architecture list` > `defaults list`. The defaults have been removed, so the next command which supports taxonomy related options will run without them.

ofedoren (changes requested):
Thanks, @yifatmakias!

Although this is pretty straightforward solution, I have one suggestion:
What if we `extend` not taxonomable resources (commands) with a `CommandExtension` which will be the same for all of them? Each command then would have something like:
```
  extend_with(HammerCLIForeman::CommandExtensions::NotTaxonomable.new)
```
Or we could add a class method `not_taxonomable` for `HammerCLIForeman::Command` that removes taxonomy related options from defaults. That could look like:
```
class ListCommand < HammerCLIForeman::ListCommand
  not_taxonomable
# ...
end
``` 
All names are just examples :)

That way we wouldn't redefine `option_sources` for this issue. Also, if there would be another problem related to that, we would need to add changes to all redefined `option_sources` instead of one place :(

lib/hammer_cli_foreman/architecture.rb
yifatmakias: @ofedoren Thanks for the review. You are definitely right and I am working on fixing it.

lib/hammer_cli_foreman/architecture.rb
```diff
@@ -11,7 +11,12 @@ class ListCommand < HammerCLIForeman::ListCommand
         field :name, _("Name")
       end
 
-      build_options
+      extend_with(HammerCLIForeman::CommandExtensions::NoTaxonomies.new)
```
ofedoren, line 14: Since this extension doesn't remove loc/org options expilictly, I'd change the name to `WithoutDefaults` or `NoDefaults`.

lib/hammer_cli_foreman/architecture.rb
```diff
@@ -11,7 +11,12 @@ class ListCommand < HammerCLIForeman::ListCommand
...
+
+      build_options do |o|
```
ofedoren, line 16: Can this be somehow moved to the extension?

yifatmakias: No, I tried it and it does not work, requires some changes in hammer core for it to work.

lib/hammer_cli_foreman/command_extensions/without_defaults.rb
```diff
@@ -0,0 +1,11 @@
+module HammerCLIForeman
+  module CommandExtensions
+    class WithoutDefaults < HammerCLI::CommandExtensions
+
+      request_options do |options|
+        options[:use_defaults] = false
```
ofedoren, line 6: Unfortunatelly this has no effect at all :(

lib/hammer_cli_foreman/architecture.rb
ofedoren: After more detailed review I've come with an idea: extension classes have access to option sources. What if we create a new option source based on `SavedDefaults` for resources that don't support taxonomies and replace `SavedDefaults` source with the new one? This way we will support default behaviour of saved defaults for commands, remove not supported options and will have code defined in one place and reused via extension.

lib/hammer_cli_foreman/bookmark.rb
```diff
@@ -15,7 +15,12 @@ class ListCommand < HammerCLIForeman::ListCommand
         field :owner_type, _('Owner Type')
       end
 
-      build_options
+      extend_with(HammerCLIForeman::CommandExtensions::WithoutDefaults.new)
+
+      build_options do |o|
```
ofedoren, line 20: With current approach this block should be added for other CRUD commands for each resource that doesn't support taxonomies. Also, those options should vanish as soon as we fix the API docs.

yifatmakias:
@ofedoren I belive I found even a better way to solve this issue then we discussed. Let me know what do you think about this solution.

ofedoren (commented):
Thanks, @yifatmakias, for digging into this even more :)

I guess this is even safer solution.

The last thing that's required is a few tests so we track this. Also, is there a way we can automatically check whether a resource is taxonomable or not? Just to make sure we didn't miss anything.

yifatmakias:
@ofedoren I added tests as we discussed. Regarding your other question, I do not think there is a way we can automatically check whether a resource is taxonomable or not but I am pretty sure that I managed to identify all the resources because I went over each and every hammer command.
Let me know if there is anything else I should fix or change :)

ofedoren:
@yifatmakias, I'd gladly merge this already, but I've got one last (I hope) request: could you please split the actual changes and rubocop changes up? Perferably into 2 PRs.

Although we don't enforce rubocop yet, I'm OK with the default settings of it. But it's out of the scope of this PR.

yifatmakias:
@ofedoren Done :) You are definitely right that the rubocop changes are out of the scope of this PR.

lib/hammer_cli_foreman/operating_system.rb
```diff
@@ -191,7 +191,7 @@ def base_action_params
         {"operatingsystem_id" => option_id}
       end
 
-      build_options
+      build_options expand: { except: %i[organizations locations] }, without: %i[organization_id location_id]
```
ofedoren, line 194: You probably want to do that for `SetOsDefaultTemplate` command as well :)

ofedoren (commented):
LGTM and ready for merge, just one last nitpick...

lib/hammer_cli_foreman/operating_system.rb
yifatmakias: Done :)

test/unit/architecture_test.rb
```diff
@@ -84,9 +92,45 @@
...
+        it_should_accept_search_params
+        it_should_fail_with 'organization param', ['--organization-id=1']
+        it_should_fail_with 'location param', ['--location-id=1']
+      end
+
+      context 'output' do
+        it 'should run list command with defaults' do
+          @api.expects_search(:users, 'login=admin').returns(index_response([user]))
+          assert_equal 2, @defaults_provider.get_defaults(:organization_id)
+          cmd.resource.call(:index)
```
ofedoren, line 125: Not sure if it actually tests anything.. I've tried to run `arch list` with foreman provider defaults and it does just one call to arch:index endpoint, no additional actions (for `host list` it does resolution for taxonomies which is correct). I guess the test should actually check that `api_expects_no_call` for getting defaults if the resource is not taxonomable (currently the test looks like it checks if architectures support defaults resolution for taxonomies, which is not right).

Also, such tests with api expectations should be moved among functional ones. The tests for params can and should be left among unit ones :)

ofedoren (commented):
Thanks, @yifatmakias, please bear with me, we're getting so close :)

I've noticed that some of the tests are not quite useful or correct, please see inline.

Otherwise LGTM :)

test/unit/architecture_test.rb
yifatmakias: I see your point but I am not sure how to write these functional tests because I don't know how can I simulate in the tests that defaults were defined if no API call accrues when setting the defaults from hammer.

I hope I managed to explain my unclarity, If not let me know.

ofedoren: @yifatmakias, yeap :) That was a little bit tricky, but a functional test for i.e. mail notifications will be:

```ruby
  it 'should run list command with defaults' do
    providers = { 'foreman' => HammerCLIForeman::Defaults.new(api_connection({}, '2.1')) }
    defaults = HammerCLI::Defaults.new(
      {
        organization_id: {
          provider: 'foreman'
        },
        location_id: {
          provider: 'foreman'
        }
      }
    )
    defaults.stubs(:write_to_file).returns(true)
    defaults.stubs(:providers).returns(providers)
    api_expects(:mail_notifications, :index, 'List mail notifications').returns(@mail_notifications)

    result = run_cmd(@cmd, { use_defaults: true, defaults: defaults })
    _(result.exit_code).must_equal HammerCLI::EX_OK
  end
```

This way we actually test that just one API call is being made and taxonomy options are ignored.

Also, it would be great to add at least one functional test for a resource that support default taxonomies, i.e. hosts:

```ruby
it 'should run list command with defaults' do
    providers = { 'foreman' => HammerCLIForeman::Defaults.new(api_connection({}, '2.1')) }
    defaults = HammerCLI::Defaults.new(
      {
        organization_id: {
          provider: 'foreman'
        },
        location_id: {
          provider: 'foreman'
        }
      }
    )
    defaults.stubs(:write_to_file).returns(true)
    defaults.stubs(:providers).returns(providers)
    api_expects(:users, :index, 'Find user').with_params(search: 'login=admin').returns(index_response([{ 'default_organization' => { 'id' => 2 } }]))
    api_expects(:users, :index, 'Find user').with_params(search: 'login=admin').returns(index_response([{ 'default_location' => { 'id' => 1 } }]))
    api_expects(:hosts, :index, 'List hosts').returns(index_response([{ 'id' => '42' }]))

    result = run_cmd(cmd, { use_defaults: true, defaults: defaults })
    _(result.exit_code).must_equal HammerCLI::EX_OK
  end
```

 This test shows that a command still works (badly optimized, but still) with taxonomy defaults.
 
 BTW, the tests could be reduced by extracting defaults/providers logic, but I'll probably leave it to you :rose:

yifatmakias: @ofedoren Thanks! I added these tests.

yifatmakias:
@ofedoren I am not sure why the tests fail, they do not fail when I run them on my environment. do you have an idea?

ofedoren (approved):
Thanks, @yifatmakias, great work! Finally merging this one :)

ekohl:
It looks like these resources aren't really taxable while the API does advertise them as such. I think the real fix is to correct the API documentation instead. Would either of you be interested in working on that?

yifatmakias:
> It looks like these resources aren't really taxable while the API does advertise them as such. I think the real fix is to correct the API documentation instead. Would either of you be interested in working on that?

I will open a redmine issue for this and work on it as soon as possible.

PR #553, opened 2020-12-16
Fixes #31570 - consume structured statuses API data

evgeni:


evgeni:
the normalize code can also be completely dropped if we only want to support foreman releases with the structured data

evgeni:
I have no idea why CI fails now, but it's not my change. (it was green before I amended the subject)

evgeni:
oh, cool, cool: https://github.com/rubygems/rubygems/issues/4229

tbrisker:
[test hammer]

shiramax (approved):

shiramax:
thanks @evgeni

shiramax:
thanks @evgeni could we have tests for this?

evgeni:
I guess so. Let me see.

evgeni:
@shiramax there, added tests for old and new behaviour, and learned something about Ruby… ;)

shiramax (approved):

shiramax:
thanks @evgeni !

PR #554, opened 2020-12-21
fixes #31536 - Removing puppetrun command

domitea:
Since the puppetrun is deprecated and removed, this PR removes command that invokes non-existing puppetrun command at Foreman side

### DIFF
--- a/lib/hammer_cli_foreman/host.rb
+++ b/lib/hammer_cli_foreman/host.rb
@@ -176,8 +176,9 @@ module HammerCLIForeman
       resource :puppet_hosts
       action :puppetrun
 
-      def print_data(records)
-        print_message _('Puppet run triggered')
+      def execute
+        warn _('The puppetrun feature has been removed, however you can use the Remote Execution Plugin to run Puppet commands')
+        HammerCLI::EX_SOFTWARE
       end
 
       build_options
--- a/test/functional/host_test.rb
+++ b/test/functional/host_test.rb
@@ -502,3 +502,17 @@ describe 'disassociate host from vm' do
     assert_cmd(expected_result, result)
   end
 end
+
+describe 'run puppetrun for host' do
+  let(:cmd) { ['host', 'puppetrun'] }
+  let(:params) { ['--id=1'] }
+
+  it "does nothing for puppetrun" do
+    expected_result = CommandExpectation.new
+    expected_result.expected_err = "The puppetrun feature has been removed, however you can use the Remote Execution Plugin to run Puppet commands\n"
+    expected_result.expected_exit_code = HammerCLI::EX_SOFTWARE
+
+    result = run_cmd(cmd)
+    assert_cmd(expected_result, result)
+  end
+end
--- a/test/unit/host_test.rb
+++ b/test/unit/host_test.rb
@@ -132,29 +132,6 @@ describe HammerCLIForeman::Host do
   end
 
 
-  context "PuppetRunCommand" do
-
-    let(:cmd) { HammerCLIForeman::Host::PuppetRunCommand.new("", ctx) }
-
-    context "parameters" do
-      it_should_accept "name", ["--name=host"]
-      it_should_accept "id", ["--id=1"]
-      # it_should_fail_with "no arguments"
-      # TODO: temporarily disabled, parameters are checked in the id resolver
-
-    end
-
-    context "output" do
-      with_params ["--id=1"] do
-        it "should inform that puppet was triggered" do
-          cmd.stubs(:context).returns(ctx.update(:adapter => :test))
-          _(proc { cmd.run(with_params) }).must_output "Puppet run triggered\n"
-        end
-      end
-    end
-  end
-
-
   context "ConfigReportsCommand" do
     before do
       ResourceMocks.mock_action_call(:config_reports, :index, [])


ofedoren (commented):
Thanks, @domitea.

Since the issue says that we need to show a deprecation message and API docs still have the endpoint marked as deprecated, I guess we can't simply remove the command right away.

What do you think?

domitea:
@ofedoren I've discussed the solution with @ares and we're think that deprecation in not necessary. The pupperrun functionality from core has been dropped already (https://github.com/theforeman/foreman/pull/7719). So this much more forgotten removal.

What do you think?

ofedoren (commented):
@domitea, well... I've taken a look at the changes if Foreman and it seems like functionallity is gone, but the endpoint with deprecation is still present. And I guess that we can remove the command when the endpoint itself is removed. For now we can show deprecation/better help message as well.

But just to be sure, @ares. You're the third, deciding, vote here :)

tbrisker:
@domitea @ofedoren @ares what's the status here? I guess we should display the same message as https://github.com/theforeman/foreman/pull/8216 in case someone tries to use this endpoint?

ofedoren:
@tbrisker, +1 for showing the same message as the server (simply use that message in hammer)

domitea:
I've updated PR with the deprecation message.

lib/hammer_cli_foreman/host.rb
```diff
@@ -173,14 +173,10 @@ def get_power_status
 
     class PuppetRunCommand < HammerCLIForeman::SingleResourceCommand
       command_name "puppetrun"
-      resource :puppet_hosts
-      action :puppetrun
```
ofedoren, line 177: Actually, whilst this is still present in API docs, there is no need to remove that. The same with `build_options`.

Oltherwise the command looks empty (which is kind of OK since we deprecate it, just looks weird).

lib/hammer_cli_foreman/host.rb
```diff
@@ -173,14 +173,10 @@ def get_power_status
...
 
-      def print_data(records)
-        print_message _('Puppet run triggered')
+      def execute
```
ofedoren, line 177: Custom `execute` method's body must explicitly end with exit code.

I'd suggest `HammerCLI::EX_OK`. If you think that the command run should end with an error, you can use simple `1`.

ofedoren (commented):
Thanks, @domitea for the update. Looks better now, but I still have few suggestions. Please, tell me what do you think.

Also, I'd suggest to adjust the test case to be simplified. The removal should be done after the whole command is removed.

domitea:
Thanks @ofedoren for review! It's a valid comments. I've updated a PR. Please take a look if it's good for you.

ares:
I think we could have dropped the command entirely, but I don't mind keeping it deprecated for a bit longer.

ofedoren:
ok to test

tbrisker:
[test hammer]

tbrisker:
Failure is due to rubygems/rubygems#4229, needing a new rubygems gem release

tbrisker:
[test hammer] looks like new bundler was released with the fix

tbrisker:
test failure related:
```
12:39:01 /home/jenkins/workspace/test_hammer_cli_foreman_pull_request/ruby/2.5/lib/hammer_cli_foreman/host.rb:177:in `<class:PuppetRunCommand>': undefined method `￼' for HammerCLIForeman::Host::PuppetRunCommand:Class (NoMethodError)
12:39:01 	from /home/jenkins/workspace/test_hammer_cli_foreman_pull_request/ruby/2.5/lib/hammer_cli_foreman/host.rb:174:in `<class:Host>'
12:39:01 	from /home/jenkins/workspace/test_hammer_cli_foreman_pull_request/ruby/2.5/lib/hammer_cli_foreman/host.rb:14:in `<module:HammerCLIForeman>'
12:39:01 	from /home/jenkins/workspace/test_hammer_cli_foreman_pull_request/ruby/2.5/lib/hammer_cli_foreman/host.rb:11:in `<top (required)>'
12:39:01 	from /home/jenkins/workspace/test_hammer_cli_foreman_pull_request/ruby/2.5/test/unit/host_test.rb:4:in `require'
12:39:01 	from /home/jenkins/workspace/test_hammer_cli_foreman_pull_request/ruby/2.5/test/unit/host_test.rb:4:in `<top (required)>'
12:39:01 	from /usr/local/rvm/gems/ruby-2.5.1@test_hammer_cli_foreman_pull_request-1/gems/rake-10.1.1/lib/rake/rake_test_loader.rb:15:in `require'
12:39:01 	from /usr/local/rvm/gems/ruby-2.5.1@test_hammer_cli_foreman_pull_request-1/gems/rake-10.1.1/lib/rake/rake_test_loader.rb:15:in `block in <main>'
12:39:01 	from /usr/local/rvm/gems/ruby-2.5.1@test_hammer_cli_foreman_pull_request-1/gems/rake-10.1.1/lib/rake/rake_test_loader.rb:4:in `select'

**Comment by domitea** [comment]
Thanks @tbrisker! There was some unusual white space that broke the test.

**Review by ofedoren — CHANGES_REQUESTED** [review]
Thanks, @domitea, it's almost done :)

One last request though: could you please move the test among functional ones? Since the test fails anyway...

Here an example you can inspire from: https://github.com/theforeman/hammer-cli-foreman/blob/a7a210a3c0f4078a2a25964268722619ddbd83bf/test/functional/host_test.rb#L23

**Comment by domitea** [comment]
@ofedoren test "moved" to functional. Is it looks good for you?

File: lib/hammer_cli_foreman/host.rb [hunk 3]
```diff
@@ -176,8 +176,9 @@ class PuppetRunCommand < HammerCLIForeman::SingleResourceCommand
       resource :puppet_hosts
       action :puppetrun
 
-      def print_data(records)
-        print_message _('Puppet run triggered')
+      def execute
+        warn _('The puppetrun feature has been removed, however you can use the Remote Execution Plugin to run Puppet commands')
+        HammerCLI::EX_OK
```
ofedoren, line 181: Thanks, @domitea! After some thoughts I think it would be better to fail this command in case if the command is being used within a script, so users won't be confused with OK status while nothing actually happened. The exit code should be probably `HammerCLI::EX_SOFTWARE`, the one that is used in case of server's 501 error.

domitea:
@ofedoren exit code updated. It's a good point and it's makes sense.

ofedoren:
@domitea, thanks :) Tests are failing. Probably we can remove those.

domitea:
@ofedoren Test removed so I hope it will pass now.

ofedoren (approved):
Thanks, @domitea! Merging now :)

PR #555, opened 2021-01-05
Fixes #8923 - Add kerberos negotiate auth support

ezr-ondrej:
Implements negotiate authentication.
Doesnt support username output.

Depends on https://github.com/Apipie/apipie-bindings/pull/81.

Expects the host hammer to be in the same REALM as Foreman server and have Foreman external authentication enabled.
User needs to manually run `kinit`, if the user is not REALM user or/and doesnt have kerberos initialization enabled by default.
If `kinit` has not been run - and thus the ticket is not in local keyring, gssapi will fail in not very user friendly manner (help wanted, but I believe this is usable already, so can be merged without it).
I do not validate foreman being able to use negotiate - it will simply fail again `^^`.

If help with testing needed, let me know.

### DIFF
--- a/lib/hammer_cli_foreman/api/negotiate_auth.rb
+++ b/lib/hammer_cli_foreman/api/negotiate_auth.rb
@@ -1,0 +1,13 @@
+module HammerCLIForeman
+  module Api
+    class NegotiateAuth < ApipieBindings::Authenticators::Negotiate
+      def initialize(foreman_url, **options)
+        super("#{foreman_url}/users/extlogin", HammerCLI::SSLOptions.new.get_options(foreman_url).merge(options))
+      end
+
+      def user
+        'UNKNOWN'
+      end
+    end
+  end
+end
--- a/lib/hammer_cli_foreman/api/negotiate_auth.rb
+++ b/lib/hammer_cli_foreman/api/negotiate_auth.rb
@@ -14,14 +14,11 @@ module HammerCLIForeman
       end
 
       def status
-        active = `klist` && $?.exitstatus.zero?
-        no_kerberos_session_msg = _('There is no active Kerberos session. Have you run %s?') % 'kinit'
-        return no_kerberos_session_msg unless active
-
-        _('No session, but there is an active Kerberos session, that will be used for negotiate login.')
-      rescue => e
-        Logging.logger['NegotiateAuth'].debug e
-        no_kerberos_session_msg
+        if system('klist')
+          _('No session, but there is an active Kerberos session, that will be used for negotiate login.')
+        else
+          _('There is no active Kerberos session. Have you run %s?') % 'kinit'
+        end
       end
 
       def error(ex)
--- a/lib/hammer_cli_foreman/api/negotiate_auth.rb
+++ b/lib/hammer_cli_foreman/api/negotiate_auth.rb
@@ -6,7 +6,18 @@ module HammerCLIForeman
       end
 
       def user
-        'UNKNOWN'
+        _('current Kerberos user')
+      end
+
+      def status
+        active = `klist` && $?.exitstatus.zero?
+        no_kerberos_session_msg = _('There is no active Kerberos session. Have you run %s?') % 'kinit'
+        return no_kerberos_session_msg unless active
+
+        _('No session, but there is an active Kerberos session, that will be used for negotiate login.')
+      rescue => e
+        Logging.logger['NegotiateAuth'].debug e
+        no_kerberos_session_msg
       end
     end
   end
--- a/lib/hammer_cli_foreman/api/negotiate_auth.rb
+++ b/lib/hammer_cli_foreman/api/negotiate_auth.rb
@@ -23,6 +23,17 @@ module HammerCLIForeman
         Logging.logger['NegotiateAuth'].debug e
         no_kerberos_session_msg
       end
+
+      def error(ex)
+        if ex.is_a?(RestClient::Unauthorized)
+          message = _('Invalid username or password.')
+          begin
+            message = JSON.parse(ex.response.body)['error']['message']
+          rescue
+          end
+          UnauthorizedError.new(message)
+        end
+      end
     end
   end
 end
--- a/lib/hammer_cli_foreman/api/negotiate_auth.rb
+++ b/lib/hammer_cli_foreman/api/negotiate_auth.rb
@@ -22,14 +22,14 @@ module HammerCLIForeman
       end
 
       def error(ex)
-        if ex.is_a?(RestClient::Unauthorized)
-          message = _('Invalid username or password.')
-          begin
-            message = JSON.parse(ex.response.body)['error']['message']
-          rescue
-          end
-          UnauthorizedError.new(message)
+        super unless ex.is_a?(RestClient::Unauthorized)
+
+        message = _('Invalid username or password.')
+        begin
+          message = JSON.parse(ex.response.body)['error']['message']
+        rescue
         end
+        UnauthorizedError.new(message)
       end
     end
   end
--- a/lib/hammer_cli_foreman/api/negotiate_auth.rb
+++ b/lib/hammer_cli_foreman/api/negotiate_auth.rb
@@ -9,6 +9,10 @@ module HammerCLIForeman
         _('current Kerberos user')
       end
 
+      def session_id
+        auth_cookie&.delete_prefix('_session_id=')
+      end
+
       def status
         active = `klist` && $?.exitstatus.zero?
         no_kerberos_session_msg = _('There is no active Kerberos session. Have you run %s?') % 'kinit'
--- a/lib/hammer_cli_foreman/exception_handler.rb
+++ b/lib/hammer_cli_foreman/exception_handler.rb
@@ -12,6 +12,7 @@ module HammerCLIForeman
         [RestClient::UnprocessableEntity, :handle_unprocessable_entity],
         [RestClient::MovedPermanently, :handle_moved_permanently],
         [RestClient::BadRequest, :handle_bad_request],
+        [ApipieBindings::AuthenticatorError, :handle_authenticator_error],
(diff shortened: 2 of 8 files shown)


shiramax:
[test hammer]

lib/hammer_cli_foreman/api/negotiate_auth.rb
```diff
@@ -0,0 +1,13 @@
+module HammerCLIForeman
+  module Api
+    class NegotiateAuth < ApipieBindings::Authenticators::Negotiate
+      def initialize(foreman_url, **options)
+        super("#{foreman_url}/users/extlogin", HammerCLI::SSLOptions.new.get_options(foreman_url).merge(options))
+      end
+
+      def user
```
shiramax, line 8: I'm getting at the end the message "Successfully logged in as 'UNKNOWN'." 
I wonder if it's the best message here or we could change  the method execute so that we see only "Successfully logged in"

shiramax (commented):
Hi @ezr-ondrej looks good!
I just have two questions: 
1. Do you think we have a way to test this? 
2.The success message might be confusing, what do you think, maybe you could change it?

lib/hammer_cli_foreman/api/negotiate_auth.rb
ezr-ondrej: oh, I forgot to comment https://github.com/theforeman/hammer-cli-foreman/pull/555/files#diff-0d4650155f0335f18fca083b6c11c7a4c9a50d4af563f44d3d003caba468c030R39

ofedoren:
Thanks, @ezr-ondrej! So sorry it took so long :(

TL;DR: I want to test https://github.com/theforeman/hammer-cli-foreman/blob/01830b5a8543d65ef0c0ac08e1d5ccca31f73aeb/lib/hammer_cli_foreman/api/session_authenticator_wrapper.rb#L40 logic to use our sessions with this type of auth.

Seems to be working and the code looks good. One thing that I've noticed while was testing:
it seems that sessions are not being used properly... Maybe I was mistaken (I need to test by hands again), but it seems that nothing such as session id is being used to re-create or re-use connection, it simly will do the steps in apipie-bindings again over and over. Am I correct?

Also, `~/.hammer/sessions/example.com` contains blank data such as `user_name` and `id`. So if I run `kinit` and `hammer auth login negotiate` I am successfully logged in. But if I run `hammer auth status` it says that I'm not logged in although I'm able to run `hammer host list` and get an answer from the server. And if I run `hammer auth logout` it will have no effect at all: no session clearing, no actual logging out (which is okay I guess since we can't run `kdestroy` within hammer)

I could make some tips for improvements IF we actually can make them, but I will need test environment again :D Few suggestions I came up with: we could try to use `_session_id` from cookie created in apipie-bindings to store it h-c-foreman sessions to re-use it again if it possible. Also, can we somehow extract `username` with GSSAPI to store it there as well?

Regarding tests: it would be nice to have them in similar manner we have them for other types (https://github.com/theforeman/hammer-cli-foreman/tree/master/test/unit/api/oauth)

ezr-ondrej:
> we could try to use `_session_id` from cookie created in apipie-bindings to store it h-c-foreman sessions to re-use it again if it possible

I think this is already happening in https://github.com/theforeman/hammer-cli-foreman/blob/01830b5a8543d65ef0c0ac08e1d5ccca31f73aeb/lib/hammer_cli_foreman/api/session_authenticator_wrapper.rb#L68 but I might be wrong and that should definitelly work, we should initiate a session and use it with the session only.

> Also, can we somehow extract `username` with GSSAPI to store it there as well?

Unfortunatelly I didn't found out how :( I could only query the user API after the session is initialized, to get this information.
That's probably why we the `auth status` doesn't show correct information :thinking:

ofedoren (commented):
@ezr-ondrej, I'm so sorry for ignoring this for such a long time, I'm finally going to merge this, but I have few questions/suggestions. Please, feel free to just put your comments/ideas/solutions and I can put them in a follow up PR so I don't block you anymore :)

I've tried this and I see few places we need to deal with:
  - Not logged in via kerberos (klist is empty):
    - hammer auth status: `Using sessions, you are currently not logged in.` **OK**
    - hammer auth login negotiate: `Error: gss_init_sec_context did not return GSS_S_COMPLETE: Unspecified GSS failure.  Minor code may provide more information
No Kerberos credentials available (default cache: KEYRING:persistent:0)` **NOT OK** (we should show a better message if we don't actually initiate kerberos auth. Is it a good idea to run `kinit` via login/password from the hammer?)
    - hammer host list: same as above.
  - Logged in via kerberos (klist shows `ofedoren`):
    - hammer auth status: `Using sessions, you are currently not logged in.` **NOT OK** (AFAIR there is no way to check that, but I've noticed that if we could run `klist` from the hammer, we could check the exit code at least: if it's `0` then there is a ticket, ideally with the right user, if it's `1` then the klist is empty.
    - _[UPD: There should be `hammer auth login negotiate` before all the below commands, but this is done automatically in [negotiate authenticator](https://github.com/Apipie/apipie-bindings/pull/81). I guess it's OK, just... feels weird_.
    - hammer host list: shows host list **OK**
    - hammer auth logout: logged out **OK**
    - hammer host list: shows host list **NOT OK** (since we tried to log out) _[UPD: That's OK since even with re-using sessions and after logging out, we simply re-login due to stored kerberos ticket, which should update session id in the session file]_
  - For both flows there were no changes in sessions file: `{"id": null, "user_name": null}`. **NOT OK**

I'm afraid that we would need to explicitly say that hammer doesn't do much when kerberos is used for auth method, it just uses primary token, which needs to be set by k* commands. Unless we actually will be calling them within hammer.

Also, if we won't be using k* commands within hammer, the user name can be something 'Current Kerberos User'. The only issue persists is that we don't store session id to re-use it again, currently all the requests are going to create a new session. I've tried to update the code, so we can store and re-use it again, but I found a one small? consequence: if we store that session id, but do `kdestroy` we will still be able to get the information by the stored session id, so it might be dangerous in case if someone will `kinit admin` and then `kdestroy`, but will forgot to run `hammer auth logout` :/ Any suggestion how to deal with it?

ofedoren:
Tests will fail until https://github.com/Apipie/apipie-bindings/pull/81 is merged and a new version of `apipie-bindings` is released.

lib/hammer_cli_foreman/exception_handler.rb
```diff
@@ -147,6 +154,24 @@ def rake_command
...
     private
 
+    def authenticator_error_message(e)
+      case e.type
+      when :negotiate then negotiation_error_message(e)
+      end
+    end
+
+    def negotiation_error_message(e)
+      case e.code
```
adamruzicka, line 164: This may not be always safe. In this particular case, try `kdestroy; hammer auth login negotiate`

```
[DEBUG 2022-05-03T11:33:40 Exception] Using exception handler HammerCLIForeman::ExceptionHandler#handle_general_exception
[ERROR 2022-05-03T11:33:40 Exception] Error: undefined method `code' for #<ApipieBindings::AuthenticatorError:0x0000559ae28ce178>
Error: undefined method `code' for #<ApipieBindings::AuthenticatorError:0x0000559ae28ce178>
[ERROR 2022-05-03T11:33:40 Exception] 

NoMethodError (undefined method `code' for #<ApipieBindings::AuthenticatorError:0x0000559ae28ce178>):
    /usr/share/gems/gems/hammer_cli_foreman-3.3.0.pre.develop/lib/hammer_cli_foreman/exception_handler.rb:165:in `negotiation_error_message'
    /usr/share/gems/gems/hammer_cli_foreman-3.3.0.pre.develop/lib/hammer_cli_foreman/exception_handler.rb:159:in `authenticator_error_message'
    /usr/share/gems/gems/hammer_cli_foreman-3.3.0.pre.develop/lib/hammer_cli_foreman/exception_handler.rb:117:in `handle_authenticator_error'
    /usr/share/gems/gems/hammer_cli-3.3.0.pre.develop/lib/hammer_cli/exception_handler.rb:31:in `handle_exception'
    /usr/share/gems/gems/hammer_cli-3.3.0.pre.develop/lib/hammer_cli/messages.rb:26:in `handle_exception'
    /usr/share/gems/gems/hammer_cli-3.3.0.pre.develop/lib/hammer_cli/abstract.rb:107:in `rescue in run'
    /usr/share/gems/gems/hammer_cli-3.3.0.pre.develop/lib/hammer_cli/abstract.rb:102:in `run'
    /usr/share/gems/gems/clamp-1.1.2/lib/clamp/subcommand/execution.rb:11:in `execute'
    /usr/share/gems/gems/clamp-1.1.2/lib/clamp/command.rb:63:in `run'
    /usr/share/gems/gems/hammer_cli-3.3.0.pre.develop/lib/hammer_cli/abstract.rb:103:in `run'
    /usr/share/gems/gems/clamp-1.1.2/lib/clamp/subcommand/execution.rb:11:in `execute'
    /usr/share/gems/gems/clamp-1.1.2/lib/clamp/command.rb:63:in `run'
    /usr/share/gems/gems/hammer_cli-3.3.0.pre.develop/lib/hammer_cli/abstract.rb:103:in `run'
    /usr/share/gems/gems/clamp-1.1.2/lib/clamp/command.rb:132:in `run'
    /usr/share/gems/gems/hammer_cli-3.3.0.pre.develop/bin/hammer:147:in `<top (required)>'
    /usr/bin/hammer:23:in `load'
    /usr/bin/hammer:23:in `<main>'
```

ofedoren: Updated both PRs, should :crossed_fingers: be fixed now.

adamruzicka (commented):
Will need an apipie-bindings dep bump once the changes there are merged and released. Apart from that, this is good to go

ofedoren:
[test hammer]

adamruzicka (approved):
:+1:

adamruzicka:
Thank you @ezr-ondrej & @ofedoren !

PR #558, opened 2021-01-25
Fixes #31630 - Global Registration module

stejskalleos:
```
bin/hammer host-registration generate-command
```


stejskalleos:
@ofedoren rebased & updated (with tests)

ofedoren (approved):
Thanks, @stejskalleos. Probably there will be no need to CP this, so it's OK to have new test data in within this commit.

LGTM and works, merging.

PR #568, opened 2021-05-08
Fixes #32454 - Show new msg for empty update commands

ofedoren:
requires https://github.com/theforeman/hammer-cli/pull/343

Most of the update commands follow API docs and put resource related params under `resource` key in `request_params`. Thanks to this, we can simply check if anything is being put there and show `Nothing to update` message if not.

### DIFF
--- a/lib/hammer_cli_foreman/command_extensions.rb
+++ b/lib/hammer_cli_foreman/command_extensions.rb
@@ -1,3 +1,4 @@
+require 'hammer_cli_foreman/command_extensions/update_common'
 require 'hammer_cli_foreman/command_extensions/fields'
 require 'hammer_cli_foreman/command_extensions/puppet_environment'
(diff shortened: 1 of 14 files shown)


yifatmakias (changes requested):
Thanks @ofedoren, but it seems that there is some issues with the tests after your changes.
Can you take a look?

ofedoren:
@yifatmakias, rebased. All the related tests are green now :)

There are 2 tests that are currently failing, but they are related to changes in https://github.com/theforeman/hammer-cli/pull/345.

I can try to fix them within this PR, but I guess it's better to create another one for that purpose.

yifatmakias (approved):
@ofedoren Thanks, looks good to me.

adiabramovitch:
@ofedoren when i am trying to use the update command I get this msg, is it just me?
Could not update the host:
  Error: undefined method `resource' for nil:NilClass
  Did you mean?  rescue

PR #570, opened 2021-05-13
Fixes #32392 - Force name resolving due defaults

ofedoren:
The issue mentions that if the `~/.hammer/defaults.yml` contains
```yaml
:defaults:
  :organization_id:
    :value: '1'
  :organization:
    :value: 'Default Organization'
```
then option `--organization-id` internally will always be set to 1 when a user specifies a different value for `--organization` option instead of one in the defaults.

This patch allows to override default `*_id` param via corresponding command line options. For example if `defaults.yml` contains `:resource_id: 1`, but on the command line `--resource 'Resource Name'` was specified and it has `id: 3`, then it will use `id: 3`.

Note: the provided `defaults.yaml` is actually a bad usage of default values for options since there is no way to choose what actually should be used in case if `organization` has id different from one mentioned as `organization_id`.

To test this PR you'll need `~/.hammer/defaults.yml` similar to provided one, but with only one of `:orgnization_id`, `organization` being set.

### DIFF
--- a/lib/hammer_cli_foreman/option_sources/id_params.rb
+++ b/lib/hammer_cli_foreman/option_sources/id_params.rb
@@ -13,22 +13,39 @@ module HammerCLIForeman
         IdParamsFilter.new(:only_required => false).for_action(@command.resource.action(@command.action))
       end
 
-      def get_options(defined_options, result)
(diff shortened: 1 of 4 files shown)


yifatmakias (approved):
Thanks @ofedoren. Looks good and works as expected.

PR #574, opened 2021-06-23
 Fixes #32861 - Prevent full-help from failure if apidoc not available

nadjaheitmann:


ofedoren (approved):
Thanks, @nadjaheitmann!

PR #575, opened 2021-06-24
Fixes #32872 - return org and loc options to all hammer commands

yifatmakias:


test/functional/architecture_test.rb
```diff
@@ -24,26 +24,6 @@
       result = run_cmd(@cmd)
       assert_cmd(expected_result, result)
     end
-
-    it 'should run list command with defaults' do
```
ofedoren, line 28: Why are you removing tests like that? I find them pretty useful. Although I'd change the version `2.1` to the latest one.

test/unit/architecture_test.rb
```diff
@@ -17,8 +17,6 @@
     context "parameters" do
       it_should_accept "no arguments"
       it_should_accept_search_params
-      it_should_fail_with 'organization param', ['--organization-id=1']
-      it_should_fail_with 'location param', ['--location-id=1']
```
ofedoren, line 21: Actually we could leave these tests as well, but test for `it_should_accept` instead :)

ofedoren (commented):
Thanks, @yifatmakias for patience and dealing with this issue! :rose:

Few inline comments though...

test/functional/architecture_test.rb
yifatmakias: Yes I agree, I think the API calls are different now but I will figure out the correct calls and fix the tests.

test/unit/architecture_test.rb
yifatmakias: Good idea I will do that.

ofedoren (approved):
Thanks, @yifatmakias, LGTM :rose:

PR #579, opened 2021-08-03
Fixes #32878 - Add parent title option for hostgroup

ofedoren:


yifatmakias (approved):
Thanks @ofedoren , looks good and works as expected.

PR #590, opened 2021-12-22
Fixes #33438 - Fix puppet-class command description

ofedoren:


adamruzicka:
Should we disregard tests when running against stable branches? The proposed changes looks fine and can't possibly cause all that breakage

ofedoren:
@adamruzicka, I guess we can simply ignore that if the tests pass locally (which they do). Such PRs into a stable branch are rare, we do CPs instead manually when we are about to release a new minor version, so we run tests locally against the current stable branch anyway.

adamruzicka (approved):
LGTM

ofedoren:
@adiabramovitch, could you please take a look/merge? :)

PR #594, opened 2022-02-16
Refs #28283 - Autodetect if apidoc cache needs reload

ofedoren:
Requires https://github.com/theforeman/hammer-cli/pull/359. Here it's enabled by default.

### DIFF
--- a/lib/hammer_cli_foreman/api/connection.rb
+++ b/lib/hammer_cli_foreman/api/connection.rb
@@ -22,8 +22,9 @@ module HammerCLIForeman
         auth_type ||= default_auth_type(settings)
         default_params = build_default_params(settings, logger, locale, auth_type)
         super(default_params,
-          :logger => logger,
-          :reload_cache => settings.get(:_params, :reload_cache) || settings.get(:reload_cache)
+          logger: logger,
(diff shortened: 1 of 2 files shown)


lib/hammer_cli_foreman/api/connection.rb
```diff
@@ -22,8 +22,9 @@ def initialize(settings, logger = nil, locale = nil, auth_type = nil)
         auth_type ||= default_auth_type(settings)
         default_params = build_default_params(settings, logger, locale, auth_type)
         super(default_params,
-          :logger => logger,
-          :reload_cache => settings.get(:_params, :reload_cache) || settings.get(:reload_cache)
+          logger: logger,
+          reload_cache: settings.get(:_params, :reload_cache) || settings.get(:reload_cache),
+          use_modules_checksum: settings.get(:foreman, :use_modules_checksum) != false
```
adamruzicka, line 27: Wouldn't this consider `nil` as truthy?

ofedoren: Yeap, and that's intentional. This is to be able to enable this by default in case the config file won't be updated by users explicitly.

ofedoren:
Closed in favor https://github.com/theforeman/hammer-cli-foreman/pull/599.

PR #601, opened 2022-05-11
Fixes #34629 - Fix config for transifex

ofedoren:
The main fix is in the first commit: fixes the config file for Transifex. It was probably a typo, but it was there since 2017 :/ And I guess nobody noticed that we didn't save pulled translations...

The second commit returns all (I hope) the currently translated strings.


adamruzicka (approved):
Good catch

adamruzicka:
Thank you @ofedoren !

PR #606, opened 2022-10-21
Fixes #35659 - ForemanGoogle plugin

stejskalleos:
For https://github.com/theforeman/foreman/pull/9272


ofedoren:
Thanks, @stejskalleos, but since the main PR is about to remove GCE as a whole, I guess we should remove the whole file here as well.

Also, there are few places to edit as well:
 - remove https://github.com/theforeman/hammer-cli-foreman/blob/master/lib/hammer_cli_foreman/compute_resource/register_compute_resources.rb#L13 line
 - GCE mentions from this https://github.com/theforeman/hammer-cli-foreman/blob/master/lib/hammer_cli_foreman/virtual_machine.rb file.

ofedoren:
@stejskalleos, https://github.com/ofedoren/hammer-cli-foreman-google has been released, so we can proceed with suggested changes here and merge them right away. The plugin can be used even with 3.4, nothing breaks :)

ofedoren (approved):
Thanks, @stejskalleos, let's get this in.

PR #611, opened 2023-04-19
Fixes #36349 - Update minitest version

archanaserver:
Adding Ruby v3.1 support.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ gem 'gettext', '>= 3.1.3', '< 4.0.0'
 group :test do
   gem 'rake', '~> 10.1.0'
   gem 'thor'
-  gem 'minitest', '5.14.1'
+  gem 'minitest', '~> 5.18'
   gem 'minitest-spec-context'
   gem 'simplecov'
   gem 'mocha', "< 1.14.0"


Gemfile
```diff
@@ -7,7 +7,7 @@ gem 'gettext', '>= 3.1.3', '< 4.0.0'
 group :test do
   gem 'rake', '~> 10.1.0'
```
ekohl, line 8: I had a quick look at if this can be updated to a current version, but it blows up.

ekohl (commented):
Our current testing pipeline doesn't run on Ruby 3 so we don't know about any regressions. IMHO that's something to address.

What I would do in the short term is just a single commit that **only** updates the minitest dependency. Nothing else. That's a good change that has immediate benefits. Adding the `required_ruby_version` part to the gemspec has more implications, like fixing CI to verify Ruby 3 as well. Should also be done, but that's no reason to hold something that is already a good thing.

evgeni:
ci config fixes in https://github.com/theforeman/jenkins-jobs/pull/331

ofedoren:
Well, since this PR is just about updating `minitest` version, can be the commit message updated? Also, even though this is a single line, I'd prefer a Redmine issue for that as well. Just in case we would need to track down the work towards Ruby 3+ support.

archanaserver:
@ofedoren the commit message is updated now, also I've created ad Redmine issue: https://projects.theforeman.org/issues/36349#change-167113
Are there any other changes that need to be made?

ofedoren:
Thanks, @archanaserver, no other changes are required :)

@evgeni, are we good to go from the packaging perspective?

ofedoren (approved):
Thanks, all :)

ofedoren:
Hmm... I've noticed that we still have even older version in hammer core: https://github.com/theforeman/hammer-cli/blob/master/Gemfile#LL10C26-L10C26. I'll open the PR.

PR #612, opened 2023-04-25
Avoid redefining methods

ekohl:
If you update rake it ends up showing more warnings, including about redefining methods. This creates a subclass for every test, avoiding the issue.

PR #613, opened 2023-04-27
Refs #36337 - Update rake for hammer-cli-foreman

ofedoren:
Replaces https://github.com/theforeman/hammer-cli-foreman/pull/612 (the commit is included).

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -10,6 +10,7 @@ Rake::TestTask.new do |t|
   t.libs.push "lib"
   t.test_files = Dir.glob('test/**/*_test.rb')
   t.verbose = true
+  t.warning = ENV.key?('RUBY_WARNINGS')
 end
 
 HammerCLI::I18n::FindTask.define(HammerCLIForeman::I18n::LocaleDomain.new, HammerCLIForeman.version)
--- a/lib/hammer_cli_foreman/commands.rb
+++ b/lib/hammer_cli_foreman/commands.rb
@@ -244,7 +244,7 @@ module HammerCLIForeman
       super
       begin
         output_definition.update_field_sets('THIN', (searchables.for(resource).map(&:name) + ['id']).map(&:to_sym))
-      rescue StandardError => e
+      rescue StandardError
         # Some subcommands may not have such fields or defined resource
       end
     end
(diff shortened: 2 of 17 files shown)


lib/hammer_cli_foreman/commands.rb
```diff
@@ -244,7 +244,7 @@ def self.output(definition = nil, &block)
       super
       begin
         output_definition.update_field_sets('THIN', (searchables.for(resource).map(&:name) + ['id']).map(&:to_sym))
-      rescue StandardError => e
+      rescue StandardError => _
```
ekohl, line 247: Isn't this the same?
```suggestion
      rescue StandardError
```

Rakefile
```diff
@@ -10,6 +10,7 @@ Rake::TestTask.new do |t|
   t.libs.push "lib"
   t.test_files = Dir.glob('test/**/*_test.rb')
   t.verbose = true
+  t.warning = false unless ENV['RUBY_WARNINGS']
```
ekohl, line 13: Why default to false? Isn't it good to see warnings?

ofedoren: It's good I guess, but most of the time we won't do anything about them (or can't due to warning coming from some library). It will also require to update any places we call `rake test` with `RUBY_WARNINGS=false` in that case.

lib/hammer_cli_foreman/commands.rb
```diff
@@ -610,8 +608,7 @@ def associated_resource
 
     def self.associated_resource(name=nil)
       @associated_api_resource = HammerCLIForeman.foreman_resource!(name) unless name.nil?
-      return @associated_api_resource if @associated_api_resource
-      return superclass.associated_resource if superclass.respond_to? :associated_resource
+      return (@associated_api_resource ||= nil) || (superclass.associated_resource if superclass.respond_to? :associated_resource)
```
ekohl, line 611: I think this is getting too complex. Isn't it better to split it over multiple lines? Though I'll admit I can't figure out what is a pretty format. The whole pattern of combining a getter and setter is tricky.

ekohl (approved):
Thanks!

PR #615, opened 2023-05-24
Add packit config

ofedoren:


ofedoren:
This should be fixed by https://github.com/theforeman/foreman-packaging/pull/9443.

adamruzicka (approved):
LGTM

@evgeni could we get your blessing here?

evgeni (approved):
this is as sane as it gets with the pre-release based rpm specs

Once I have a better solution via https://github.com/Katello/katello/pull/10566 I'll post a PR here too

adamruzicka:
Thank you @ofedoren & @evgeni !

PR #620, opened 2023-11-02
i18n - extracting new, pulling from tx

ofedoren:


adamruzicka (commented):
Huh, I would have expected that the removal of rackspace would have some impact on the strings that get extract, but apparently not. The only things that seem to be changed are metadata and comments in the pot file. Do we want to get this in even if the changes most likely won't be visible to the users?

ofedoren:
I'd say why not? :) If it doesn't break anything though.

adamruzicka (approved):
Alright, lgtm

adamruzicka:
Thank you @ofedoren

PR #622, opened 2023-11-29
Fixes #36954 - add cache status to ping output

evgeni:


lib/hammer_cli_foreman/ping.rb
```diff
@@ -12,6 +12,12 @@ class ForemanCommand < HammerCLIForeman::Command
             field :active, _('Status')
             field :duration_ms, _('Server Response')
           end
+          field :cache, _('cache'), Fields::Label do
+            field :servers, _('servers'), Fields::Collection do
+              field :status, _('Status')
+              field :duration_ms, _('Server Response')
```
evgeni, line 18: Should this be "server response time", or should we add some translation like we have in 
https://github.com/theforeman/hammer-cli-foreman/blob/2c191555979732a4dcd395337a11c662943b603a/lib/hammer_cli_foreman/command_extensions/ping.rb#L9

ofedoren (approved):
Thanks, @evgeni, I'm going to release new major version today and wanted to include this change as well, thus I've pushed additional commit just to make thing more consistent and at the same time answering your question :)

PR #623, opened 2023-11-29
use strings, not floats to denote ruby versions

evgeni:
otherwise `3.0` is translated to `3` which means something different :)


ofedoren (approved):
Ah, damn... Thanks, @evgeni !

PR #626, opened 2024-02-14
Fixes #37163 - Use require_relative for 'coverage_reporter'

nofaralfasi:
Replace the usage of require with require_relative for 'coverage_reporter', ensuring that it is loaded relative to the current file's directory.


ofedoren:
Thanks, @nofaralfasi, could we please track this via Redmine ticket?

ofedoren (approved):
Thanks a lot, @nofaralfasi !

PR #630, opened 2024-07-25
Fixes #37685 - Add Secure Boot and TPM options to VMware

nofaralfasi:
- Add `uefi_secure_boot` firmware for UEFI with Secure Boot enabled.
- Add `virtual_tpm` option to compute-attributes.

Requires:
- https://github.com/theforeman/foreman/pull/10324
- https://github.com/fog/fog-vsphere/pull/305


ofedoren (approved):
Thanks, @nofaralfasi !

PR #638, opened 2025-03-14
Fixes #38291 - fix hammer host update error messages

jeremylenz:
This should fix https://projects.theforeman.org/issues/38253 and a few other issues with `hammer host update`.

The Foreman exception handler doesn't look at `displayMessage` in the response, but that's where the error message is.

So with this change, instead of 
```
Could not update the host:
  500 Internal Server Error
```

you'll get the actual error message (for example):

```
Could not update the host:
  RuntimeError: Cannot assign content view environment elephant/centos_old: The content view has either not been published or has not been promoted to that lifecycle environment.
```

### DIFF
--- a/lib/hammer_cli_foreman/exception_handler.rb
+++ b/lib/hammer_cli_foreman/exception_handler.rb
@@ -98,9 +98,11 @@ module HammerCLIForeman
       begin
         response = JSON.parse(e.response)
         response = HammerCLIForeman.record_to_common_format(response) unless response.has_key?('message')
-        message = response['message'] || e.message
+        message = response["displayMessage"] || response["full_messages"] || response["message"]
       rescue JSON::ParserError
         message = e.message
+      ensure
+        message ||= e.message
       end
 
       print_error message


qcjames53 (approved):
This was surprisingly tricky to test, but I've been able to observe the correct display of error messages on a wide range of host update failures, confirming that the root issue has been addressed. Every other error type I've tested appears to be identical to the previous messages. I did not run across a response where displayMessage or full_messages taking precedence over the other message types ended up being an issue. If anything I missed is broken, it should be a real quick fix in the future. ACKing.

ehelms:
Is `RuntimeError` apprpriate? or normal for this type of validation error coming from the server?

jeremylenz:
> Is `RuntimeError` apprpriate? or normal for this type of validation error coming from the server?

I can fix that in https://github.com/Katello/katello/pull/11338

jeremylenz:
@ehelms Updated the [Katello PR](https://github.com/Katello/katello/pull/11338/files).

ehelms:
> @ehelms Updated the [Katello PR](https://github.com/Katello/katello/pull/11338/files).

What kind of error does CLI report now?

qcjames53:
I'm on it :+1:

qcjames53:
@ehelms Here's what we're seeing now:
```
[vagrant@centos9-hammer-rpm hammer-cli]$ bundle exec hammer host create --location-id 2 --organization-id 1 --name centos9-host-3 --mac 52:54:00:10:34:dc --domain-id 1 --operatingsystem-id 1 --architecture-id 1 --root-password changeme --partition-table-id 142 --medium-id 3 --lifecycle-environment-id 3 --content-view-id 1
Could not create the host:
  Cannot assign content view environment RHEL 9 LE/Default Organization View: The content view has either not been published or has not been promoted to that lifecycle environment.
  ```

**Comment by jeremylenz** [comment]
Interesting, so if I pass in a specific error class to `fail`, the error class name no longer ends up in Hammer output. Well, I'm happy with that if you are!

**Comment by qcjames53** [comment]
Just to clarify, the error type names still render for some of the more obscure errors, just not that one anymore. It follows the style of messages like having a bad org id and such. I'm okay with this if you are.

Example of bad medium id:
```
Could not create the host:
  PG::ForeignKeyViolation: ERROR:  insert or update on table "hosts" violates foreign key constraint "hosts_medium_id_fk"
  DETAIL:  Key (medium_id)=(100) is not present in table "media".
 ```

ehelms:
The output from https://github.com/theforeman/hammer-cli-foreman/pull/638#issuecomment-2743469353 looks good.

The output from https://github.com/theforeman/hammer-cli-foreman/pull/638#issuecomment-2743485378 is displaying internal information that is not relevant to a CLI user. This output comes from the server and would need to be handled more cleanly there.

PR #641, opened 2025-05-28
Fixes #TBFL - Ldap group membership

adamruzicka:
Draft until the main PR settles

### DIFF
--- a/lib/hammer_cli_foreman/auth_source_ldap.rb
+++ b/lib/hammer_cli_foreman/auth_source_ldap.rb
@@ -30,7 +30,7 @@ module HammerCLIForeman
           field :account, _('Account Username')
           field :base_dn, _('Base DN')
           field :groups_base, _('Groups base DN')
-          field :use_netgroups, _('Use Netgroups'), Fields::Boolean
+          field :ldap_group_membership, _('LDAP group membership')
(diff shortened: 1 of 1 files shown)


lib/hammer_cli_foreman/auth_source_ldap.rb
```diff
@@ -30,7 +30,7 @@ class InfoCommand < HammerCLIForeman::InfoCommand
           field :account, _('Account Username')
           field :base_dn, _('Base DN')
           field :groups_base, _('Groups base DN')
-          field :use_netgroups, _('Use Netgroups'), Fields::Boolean
```
ofedoren, line 33: This is still present in the response, should we rather just add `ldap_group_membership`?

We also can mark `use_netgroups` as deprecated in case we're going to drop that:

```ruby
field :ldap_group_membership, _('LDAP group membership')
field :use_netgroups, _('Use Netgroups'), Fields::Boolean, :replaced_by => [_('Account'), _('LDAP group membership')].join('/')
```

ofedoren (commented):
Thanks, @adamruzicka, please create a RM ticket and link it to the one from Foreman so we better track the whole effort.

Some suggestions inline:

ofedoren (approved):
Thanks, @adamruzicka !

PR #658, opened 2026-07-17
Fixes #39537 - Implement LDAP CA cert support

adamlazik1:
Cacert can be provided as file or string.

Made with Cursor.

Requires https://github.com/theforeman/foreman/pull/11107

### DIFF
--- a/lib/hammer_cli_foreman/auth_source_ldap.rb
+++ b/lib/hammer_cli_foreman/auth_source_ldap.rb
@@ -7,14 +7,9 @@ module HammerCLIForeman
     module CacertFileOption
       def self.included(base)
         base.option '--cacert-file', 'CACERT_FILE',
-                    _('Path to a PEM file containing CA certificate(s) for LDAPS verification. Ignored if --cacert is set'),
+                    _('Path to a PEM file with CA certificate(s) added to the system trust store for LDAPS verification'),
+                    :attribute_name => :option_cacert,
                     :format => HammerCLI::Options::Normalizers::File.new
-      end
-
-      def request_params
-        params = super
-        params['auth_source_ldap']['cacert'] ||= options['option_cacert_file'] if options['option_cacert_file']
-        params
       end
     end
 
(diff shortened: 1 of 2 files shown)


lib/hammer_cli_foreman/auth_source_ldap.rb
```diff
@@ -4,6 +4,20 @@ class AuthSourceLdap < HammerCLIForeman::Command
...
+    module CacertFileOption
+      def self.included(base)
+        base.option '--cacert-file', 'CACERT_FILE',
+                    _('Path to a PEM file containing CA certificate(s) for LDAPS verification. Ignored if --cacert is set'),
+                    :format => HammerCLI::Options::Normalizers::File.new
+      end
+
+      def request_params
+        params = super
+        params['auth_source_ldap']['cacert'] ||= options['option_cacert_file'] if options['option_cacert_file']
```
adamruzicka: I'm not exactly happy about one option silently shadowing the other. An sort of established pattern is to make hammer hide the original option (see [h-cli-rex](https://github.com/theforeman/hammer_cli_foreman_remote_execution/blob/master/lib/hammer_cli_foreman_remote_execution/job_template.rb#L59) for reference). I'd prefer either that or making the two options mutually exclusive explicitly.

adamlazik1: Went with the first approach.

adamruzicka (commented):
Will need to test this, but at a glance it looks good.

adamlazik1:
For now, I tested it with https://github.com/theforeman/foreman/pull/11107.

lhellebr:
ACK as per https://github.com/theforeman/foreman/pull/11107#issuecomment-5142155899

I've used a packit build in all scenarios since Hammer is not containerized

adamruzicka:
We can undraft now

adamruzicka (approved):

adamruzicka:
Thank you @adamlazik1 & @lhellebr !

